PR #14, opened 2020-06-26
Ensure buffer isn't exhausted before processing Options

Chris-Corea:


koap/src/main/kotlin/Decoder.kt
```diff
@@ -206,13 +206,13 @@ fun ByteArray.decodeTcp(): Message.Tcp {
...
     // +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-    val options = content.readOptions()
+    val options = if (!content.exhausted()) content.readOptions() else emptyList()
 
     // |7 6 5 4 3 2 1 0|7 6 5 4 3 2 1 0|7 6 5 4 3 2 1 0|
     // +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
     // |    Payload (if any) ...
     // +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-    val payload = content.readByteArray()
+    val payload = if (!content.exhausted()) content.readByteArray() else byteArrayOf()
```
twyatt: The `exhausted` check is not actually needed here, as `readByteArray` will return an empty `ByteArray` if there are no more bytes to read.

```suggestion
    val payload = content.readByteArray()
```

twyatt:
Released to [Maven Central as `0.2.2`](https://search.maven.org/artifact/com.juul.koap/koap/0.2.2/jar):

```groovy
repositories {
    jcenter() // or mavenCentral()
}

dependencies {
    implementation 'com.juul.koap:koap:0.2.2'
}
```

PR #16, opened 2020-06-26
Use Shields.io for Maven Central badge

twyatt:
Previous badge was not updating with the latest version.


twyatt:
Closing for now, as the existing badge actually does seem to update (it just seems to take a _very_ long time; on the order of 12+ hours).

PR #17, opened 2020-06-27
Support decoding header independent of message

twyatt:
The CoAP header can now be decoded independent of the entire encoded message. Useful in situations where the entire message is only needed conditionally based on information available in the header. Usage documentation available under the [README's **Header** section](https://github.com/JuulLabs/koap/tree/twyatt/decode-header#header).

Introduces `ByteArrayReader` that wraps (and reads) from a backing `ByteArray` (which avoids the [extra copy imposed by Okio](https://stackoverflow.com/a/54545370)).

The `token` was also changed to be non-`null`, as per [RFC 7252 (pg. 33) 5.3.1. Token](https://tools.ietf.org/html/rfc7252#section-5.3.1):

> The Token is used to match a response with a request. The token value is a sequence of 0 to 8 bytes. **(Note that every message carries a token, even if it is of zero length.)** Every request carries a client-generated token that the server MUST echo (without modification) in any resulting response.


twyatt:
Snapshot available as:

```groovy
repositories {
}

dependencies {
    implementation "com.juul.koap:koap:0.3.0-SNAPSHOT"
}
```

Chris-Corea:
> Note that every message carries a token, even if it is of zero length.

Oof. Can you help me understand how a message would carry a token if the token is of zero length? What does this mean for parsing messages? Does that mean the token byte is still populated with a zero value?

koap/src/main/kotlin/ByteArrayReader.kt
```diff
@@ -0,0 +1,84 @@
...
+        check(index < endIndex) { "Cannot read at index $index, >= endIndex of $endIndex" }
+    }
+
+    /** Reads 1-byte to acquire an 8-bit unsigned int. */
+    fun readUByte(): Int {
+        checkIndex()
+        return bytes[index++].toInt() and 0xFF
+    }
+
+    /** Reads 2-bytes in [ByteArray] receiver to acquire a 16-bit unsigned int. */
```
Chris-Corea, line 39: > in [ByteArray] receiver

nit: redundant 🤷‍♂️

twyatt:
> > Note that every message carries a token, even if it is of zero length.
> 
> Oof. Can you help me understand how a message would carry a token if the token is of zero length? What does this mean for parsing messages? Does that mean the token byte is still populated with a zero value?

I took it to mean that a zero-length token should be assumed to be a `token` value of `0`.

Which makes some sense, in that it describes when writing the token you should begin writing at the first non-zero byte. If your token value is `0`, then you'll never write any token bytes.

Additionally, you'll never need to have a byte to represent `0`, as the absence of that byte (with a value `0x00`) means `token` value of `0`.

koap/src/main/kotlin/ByteArrayReader.kt
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    /** Reads 4-bytes in [ByteArray] receiver to acquire a 32-bit unsigned int. */
+    fun readUInt(): Long {
+        checkIndex()
+        return ((bytes[index++].toLong() and 0xFF) shl 24) or readUInt24().toLong()
+    }
+
+    /** Reads 8-bytes in [ByteArray] receiver to acquire a 64-bit signed int. */
+    fun readLong(): Long {
```
Chris-Corea, line 58: No change request with this comment, just want to get your thoughts on the following:

Do you think it's worth checking for the specific number of bytes you expect to read? Or is that a user problem?

In my opinion, I think it would be a nice-to-have if this class threw its own exception stating that there aren't enough bytes to read for the requested read operation. I don't think it would provide a huge gain for the user over seeing an `IndexOutOfBoundsException`, but I think it would be good to have our own sanity check to throw a specific exception per read operation if there aren't enough bytes to read.

koap/src/test/kotlin/DecoderTest.kt
```diff
@@ -139,15 +139,34 @@ class DecoderTest {
...
+            code = GET,
+            token = 0,
+            options = emptyList(),
+            payload = byteArrayOf(0x01, 0x02, 0x03)
+        )
+        val extraData = byteArrayOf(0x0A, 0x0B, 0x0C)
+
+        assertEquals(
+            expected = message,
+            actual = (message.encode() + extraData).decode()
```
Chris-Corea, line 155: Nice 💯

koap/src/main/kotlin/ByteArrayReader.kt
```diff
@@ -0,0 +1,84 @@
+package com.juul.koap
+
+internal inline fun <T> ByteArray.withReader(
+    startIndex: Int = 0,
+    endIndex: Int = size,
+    action: ByteArrayReader.() -> T
+): T = action.invoke(ByteArrayReader(this, startIndex, endIndex))
```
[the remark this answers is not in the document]
twyatt: Using a higher function to provide DSL style invocation is common, though this isn't a DSL, and I'm not aware of a standard of naming of variables holding a high-order function.

Ultimately I was just lazy with internal API docs. A KDoc here is certainly warranted so I'll add it. :+1:

koap/src/main/kotlin/ByteArrayReader.kt
twyatt: Oh, old doc, no more receiver even. Good catch.

koap/src/main/kotlin/ByteArrayReader.kt
twyatt: Definitely agree. We should provide informative exceptions. That's what `checkIndex` does, but the exception message is a bit cryptic, so I'll fix that.

twyatt: Updated `checkIndex` to hopefully be more readable?

```kotlin
private fun checkIndex() {
    if (index >= endIndex) {
        throw IndexOutOfBoundsException(
            "Cannot read when index is at or beyond endIndex (index=$index, endIndex=$endIndex)"
        )
    }
}
```

twyatt:
> `ByteArrayReader` is rad 👍
> 
> I propose moving this into `tuulbox`!

`tuulbox` is currently internal-only so I'll move it if we find a need to use it in another library; until then it's nice to leave it in this open-source library. 😄

koap/src/main/kotlin/ByteArrayReader.kt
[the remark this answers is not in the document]
Chris-Corea: Oh sorry, duh. Didn't realize this was `internal`.

Chris-Corea:
Just a thought after attempting to integrate this into tin-can: It would be nice if the header objects could provide the raw bytes the header was built from. 

At least for tin-can, I need the individual header bytes and options/payload bytes for the encryption step (since only the payload is actually encrypted and the header data is used as additional data for the encryption step). As such it would be great if I could do something like:

```kotlin
val message = Message.Udp(...)

val udpHeaderBytes = message.encode().headerBytes()
val payloadBytes = message.encode().payloadBytes()
```

Or maybe there could be a way to get the message header size and payload size from the `Message` object?
```kotlin
val message = Message.Udp(...)
val headerSize = message.headerSize // or headerSize?
val payloadSize = message.payloadSize // or payloadSize

val messageBytes = message.encode()
val udpHeaderBytes = messageBytes.copyOf(0, headerSize)
val payloadBytes = ...
```

koap/src/main/kotlin/Decoder.kt
```diff
@@ -84,15 +98,125 @@ inline fun <reified T : Message> ByteArray.decode(): T =
...
+ * decoding after header, for example:
+ *
+ * ```
+ * val header = encoded.decodeUdpHeader()
+ * val message = encoded.decode(header, offset = header.size)
+ * ```
+ */
+fun ByteArray.decode(
+    header: Header.Udp,
+    offset: Int = 0
```
cedrickcooke, line 142: Would it make more sense to remove the default? Or possibly have the default as `offset: Int = header.size`?

koap/src/main/kotlin/Decoder.kt
```diff
@@ -84,15 +98,125 @@ inline fun <reified T : Message> ByteArray.decode(): T =
...
+ * decoding after header, for example:
+ *
+ * ```
+ * val header = encoded.decodeTcpHeader()
+ * val message = encoded.decode(header, offset = header.size)
+ * ```
+ */
+fun ByteArray.decode(
+    header: Header.Tcp,
+    offset: Int = 0
```
cedrickcooke, line 159: Same here

koap/src/main/kotlin/Decoder.kt
```diff
@@ -351,8 +429,25 @@ private fun Byte.toCode(): Code = when (val code = toInt()) {
...
+ */
+internal fun ByteArrayReader.readNumberOfLength(
+    bytes: Int
+): Long = when (bytes) {
+    0 -> 0L
+    1 -> readUByte().toLong()
+    2 -> readUShort().toLong()
+    3 -> readUInt24().toLong()
+    4 -> readUInt()
+    else -> throw IllegalArgumentException("Unsupported number length of $bytes bytes")
```
cedrickcooke, line 452: Should there also be an `8 -> readLong()`? Not necessarily that we use it, just seems odd to skip one of the available functions.

koap/src/main/kotlin/ByteArrayReader.kt
```diff
@@ -0,0 +1,89 @@
...
+    fun readUInt24(): Int {
+        checkIndex()
+        return ((bytes[index++].toInt() and 0xFF) shl 16) or readUShort()
+    }
+
+    /** Reads 4-bytes to acquire a 32-bit unsigned int. */
+    fun readUInt(): Long {
+        checkIndex()
+        return ((bytes[index++].toLong() and 0xFF) shl 24) or readUInt24().toLong()
+    }
```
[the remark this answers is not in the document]
twyatt: Indeed! #18
😄

koap/src/main/kotlin/Constants.kt
```diff
@@ -9,4 +9,26 @@ internal const val COAP_VERSION = 1
...
+internal const val USHORT_MAX_VALUE = 65_535 // 2^16-1
+
+/** 16-bit unsigned integer (UShort) range. */
+internal val USHORT_RANGE = 0..USHORT_MAX_VALUE
+
+/** 32-bit unsigned integer (UInt) maximum value. */
+internal const val UINT_MAX_VALUE = 4_294_967_295 // 2^32-1
+
+/** 32-bit unsigned integer (UInt) range. */
+internal val UINT_RANGE = 0..UINT_MAX_VALUE
```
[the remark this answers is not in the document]
twyatt: #18

koap/src/main/kotlin/Decoder.kt
twyatt: I was really torn on this one.
Couldn't decide what would be more commonly used.

On one hand, if you've decoded the header, it's probably that you've discarded the header data and are now only holding onto the "content" (Options + Payload) of the CoAP encoded message. In this case, `offset` of `0` makes the most sense.

I think from an API perspective though, a default of `header.size` makes the most sense (is the most intuitive) when glancing at this function signature.

I'll make the change, thanks for the suggestion!

koap/src/main/kotlin/Decoder.kt
twyatt: The `8` (bytes) is absent here because there are no mentions of "64-bit" in the RFC in re: to encoding/decoding.
As far as I know, the larges supported unsigned integer size specified by the RFC is 32-bit; which is good since I opted out of using the experimental unsigned integers in Kotlin (we'd have a lot of potential overflows to deal with if the RFC specified 64-bit **unsigned** integers.

twyatt: I overlooked that token is being read from 8 bytes (albeit as a **signed** integer).
@cedrickcooke you were correct, you tried to prevent a regression but I didn't listen. Oops
Fixed in #20.

PR #19, opened 2020-06-30
Add ability to encode only the header of a UDP message

twyatt:


twyatt:
Snapshot available as:

```groovy
repositories {
}

dependencies {
    implementation "com.juul.koap:koap:0.3.0-SNAPSHOT"
}
```

Chris-Corea:
> Does COAP have a HEAD method like HTTP? I am struggling to see the use for this. But still, no harm in having it.

I don't think there was anything about a HEAD method in the RFC (but I could be wrong, that thing was basically a wall of text). 

The need to encode only the header of a UDP message is most likely specific to JUUL needs, as the header is required for use during the encryption of secure coap messages. I'm working on that part of tin-can at the moment so hopefully the pieces will start to come together once you see that PR.

PR #20, opened 2020-07-01
Fix reading tokens of length 8 (bytes)

twyatt:
#17 introduced a regression whereas a token of length 8-bytes would throw an `IllegalArgumentException`. Was an oversight that despite the RFC not stating a need to read 64-bit numbers, we store the token in a `Long` (since it can be up to 8-bytes in length).


twyatt:
@Chris-Corea can you validate that this works as expected?

```groovy
repositories {
}

dependencies {
    implementation "com.juul.koap:koap:0.3.1-SNAPSHOT"
}
```

Once validated, I'll cut a release.

Chris-Corea:
@twyatt Confirmed it works in a unit test in `tin-can`. Thanks for the quick turnaround!

PR #27, opened 2020-10-09
[APD-954] Publish to NPM

davidtaylor-juul:
Adds a new `publishNpm ` Gradle task (from npm.gradle) based mostly on one borrowed from the coroutines repo with some hackage.

This task expects the Webpack output from `jsBrowserProductionWebpack` task to already exist.

The `publishNpm` task creates the `npmdist` output folder, copies in the Webpack artifacts, and uses `<module>/npm/package.json` as a template to create a new `package.json` file. It then runs the npm publish command from this folder which publishes the package on GitHub in the `@Juullabs` scope.

When reviewing please provide guidance on best practices on the Gradle stuff. Thus far the effort has been just getting it to work and it would be nice to apply some polish to any areas that are crunchy.

*WIP: Currently specifies `dryRun=true` and a deliberate syntax error (`notpublish` instead of `publish`) to avoid spamming public packages while testing*

### DIFF
--- a/koap/build.gradle
+++ b/koap/build.gradle
@@ -39,10 +39,15 @@ kotlin {
         apply plugin: 'kotlinx-serialization'
         apply from: rootProject.file('gradle/npm.gradle')
 
+        //moduleName = "koap-core"
+
         browser {
-            // distribution {
-            //     directory = file("$projectDir/mydist/")
-            // }
+            distribution {
+                directory = file("$buildDir/npmdist/")
+            }
+            webpackTask {
+                // ?
+            }
         }
 
         binaries.executable()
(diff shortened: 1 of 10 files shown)


koap/build.gradle
```diff
@@ -37,7 +37,20 @@ kotlin {
 
     js {
         apply plugin: 'kotlinx-serialization'
-        browser()
+        apply from: rootProject.file('gradle/npm.gradle')
+
+        //moduleName = "koap-core"
```
twyatt, line 42: Does this do anything, or can we just remove it?

twyatt:
When you have a chance, please rebased onto `twyatt/js`.
Unless this was intended more as a proof-of-concept, in which case you can leave it as-is.

PR #30, opened 2020-10-15
[APD-954] Publish to NPM

davidtaylor-juul:
This is a re-hack of #27 

Adds a new `publishNpm ` Gradle task (from npm.gradle) based mostly on one borrowed from the coroutines repo with some hackage.

This task expects the Webpack output from `jsBrowserProductionWebpack` task to already exist.

The `publishNpm` task creates the `npmdist` output folder, copies in the Webpack artifacts, and uses `<module>/npm/package.json` as a template to create a new `package.json` file. It then runs the npm publish command from this folder which publishes the package on GitHub in the `@Juullabs` scope.

When reviewing please provide guidance on best practices on the Gradle stuff. Thus far the effort has been just getting it to work and it would be nice to apply some polish to any areas that are crunchy.

*WIP: Currently specifies `dryRun=true` to avoid spamming public packages while testing*


gradle/npm.gradle
```diff
@@ -0,0 +1,105 @@
...
+
+def authToken = prop("kotlin.npmjs.auth.token", "")
+def dryRun = prop("dryRun", "true")
+
+
+ext {
+    // These are referenced from the package.json template
+    kotlinVersion = "1.4.10"
+    npmPublishVersion = prop("npmPublishVersion", getLatestGitTag())
+}
```
davidtaylor-juul, line 47: These are not actually used. Anything defined outside of a task is no longer available when using the Kotlin DSL. I don't know how to resolve this problem.

gradle/npm.gradle
```diff
@@ -0,0 +1,105 @@
...
+// TODO: add dependsOn: [jsBrowserProduction/DevelopmentWebpack]
+task preparePublishNpm(type: Copy) {
+    def npmTemplateDir = file("$projectDir/npm")
+    def npmDeployDir = file("$buildDir/npmdist")
+    def kotlinVersion = "1.4.10"
+
+    // This no longer works....?
+    // def outputs = kotlin.targets.js.compilations.main.output.allOutputs
+    // TODO: probably instead should get the value of jsBrowserProduction/DevelopmentWebpack task outputs...
+    def outputs = file("$buildDir/distributions")
```
davidtaylor-juul, line 59: Using `main.output.allOutputs` used to work prior to the Kotlin DSL change. I can't find what to use now...

gradle/npm.gradle
```diff
@@ -0,0 +1,105 @@
...
+
+    // println("jsLegacy: $kotlin.targets.jsLegacy.compilations.main.output.allOutputs")
+    from(npmTemplateDir) {
+        // Postpone expansion of package.json until we configure version property in build.gradle
+        include "package.json"
```
davidtaylor-juul, line 64: The copy here used to perform variable substitution in the `package.json` file. This does not work with Kotlin DSL. I also tried the below method to use `copySpec.expand` but that also does not work (perhaps I am doing it wrong)?

Either way this is exacerbated by the variables setup during configuration not being available anymore.

koap/build.gradle.kts
```diff
@@ -26,7 +27,16 @@ tasks.withType<JacocoReport> {
...
-    js().browser()
+    js {
+        apply(from = "${rootDir}/gradle/npm.gradle")
+        browser {
+            distribution {
+                file("${buildDir}/npmdist/")
+            }
+        }
+        binaries.executable()
+        // target = "v5"
```
davidtaylor-juul, line 38: I haven't been able to figure out the syntax to specify the target java version here...

koap/npm/package.json
```diff
@@ -0,0 +1,24 @@
+{
+    "name": "@juullabs/koap",
+    "version" : "0.3.2-test1",
```
davidtaylor-juul, line 3: Hard coded version as cannot get substitution to work...this used to be simply `"$version"`

koap/npm/package.json
```diff
@@ -0,0 +1,24 @@
...
+      "type": "git",
+      "url": "git+https://github.com/JuulLabs/koap.git"
+    },
+    "keywords": [
+      "Kotlin",
+      "CoAP",
+      "JavaScript"
+    ],
+    "peerDependencies": {
+        "kotlin": "^1.4.2"
```
davidtaylor-juul, line 22: More hardcoded versions...

koap/build.gradle.kts
twyatt: > target java version

I assume you meant **javascript**.

We can specify the target Java version via (this can go at the top-level of this file, i.e. does not need to be under `kotlin { jvm { ... } }`):

```kotlin
tasks.withType<KotlinCompile>().all {
    kotlinOptions {
        jvmTarget = "1.8"
    }
}
```

Maybe worth trying a variant of the above snippet and seeing if there is a similar Javascript version?

twyatt (commented):
It's unfortunate that so much functionality broke when moving to Kotlin DSL (at least for the components that remained in Groovy Gradle land), so, apologies there, did not realize the transition would mess with your efforts so much.

I suppose it's better to make the Koltin Gradle transition sooner than later though, as we don't want to build a bunch of Groovy Gradle tooling if that is going away. ¯\\_(ツ)\_/¯

---

Perhaps we should try migrating `npm.gradle` to Kotlin Gradle (`npm.gradle.kts`)? I'm not sure you'd hit any less road blocks, but at least blockers you hit are less of a mystery to me (since there are often "clever" / creative Kotlin-powered workarounds that we can try); whereas I'm clueless in the Groovy Gradle realm and can't offer much help at all.

koap/build.gradle.kts
davidtaylor-juul: Sorry of the confusion, I mis-typed I meant javascript not java ☕ 

I was trying to specify the ECMA target as specified here:
https://kotlinlang.org/docs/reference/using-gradle.html#attributes-specific-for-js

davidtaylor-juul:
> Perhaps we should try migrating `npm.gradle` to Kotlin Gradle (`npm.gradle.kts`)?

I will try this out. And I am also going to revisit the other `npm-publish` plugin I tried that is in Kotlin DSL and is incompatible with the groovy Gradle.

koap/build.gradle.kts
twyatt: We can pass compiler arguments to the Kotlin compiler. So we'll probably have to figure out what the previous `target = "v5"` was resolving to as far as a compiler arg.

PR #31, opened 2020-10-19
APD-954: Publish to NPM

davidtaylor-juul:
Another NPM publish attempt, this time using the plugin from https://gitlab.com/lt.petuska/npm-publish

Replaces PR #30 

This plugin adds the tasks `assembleJsNpmPublication` and `publishJsNpmPublicationToGithub`

## Example
```
% ./gradlew clean publishJsNpmPublicationToGithub -Pnpm.publish.dry=true -Pnpm.publish.authToken.github=fakingit -Pnpm.publish.version=v0.0.3-test19

> Task :koap:publishJsNpmPublicationToGithub
npm notice 
npm notice 📦  @juullabs/koap@0.0.3-test19
npm notice === Tarball Contents === 
npm notice 109.5kB koap-koap.js                     
npm notice 4.7kB   koap-koap.meta.js                
npm notice 338B    package.json                     
npm notice 9.2kB   koap-koap/com/juul/koap/koap.kjsm
npm notice 1.0MB   koap-koap.js.map                 
npm notice === Bundled Dependencies === 
npm notice kotlin-test-js-runner
npm notice kotlin-test
npm notice kotlin
npm notice okio-okio
npm notice === Tarball Details === 
npm notice name:          @juullabs/koap                          
npm notice version:       0.0.3-test19                            
npm notice package size:  1.0 MB                                  
npm notice unpacked size: 9.8 MB                                  
npm notice shasum:        0d8b9860bf28bbea8c64dfec08a86552db17046b
npm notice integrity:     sha512-z7H8WcDT0REBA[...]PgBiopj+gznvA==
npm notice bundled deps:  4                                       
npm notice bundled files: 20                                      
npm notice own files:     5                                       
npm notice total files:   25                                      
npm notice 
+ @juullabs/koap@0.0.3-test19
```

package.json:
```js
{
  "name": "@juullabs/koap",
  "version": "v0.0.3-test19",
  "main": "koap-koap.js",
  "dependencies": {
    "okio-okio": "2.9.0",
    "kotlin": "1.4.10",
    "kotlin-test-js-runner": "1.4.10",
    "kotlin-test": "1.4.10"
  },
  "bundledDependencies": [
    "okio-okio",
    "kotlin",
    "kotlin-test-js-runner",
    "kotlin-test"
  ]
}
```


koap/build.gradle.kts
```diff
@@ -58,3 +62,53 @@ kotlin {
...
+        publication("${name}") {
+            moduleName = "${name}"
+            main = "${name}.js"
+
+            files {
+                from("${buildDir}/distributions")
+            }
+
+            dependencies {
+                npmPeer("kotlin", "*")
```
davidtaylor-juul, line 104: This is kind of a workaround for being unable to forward on the Kotlin compiler/library version - asterisk means it can be whatever.

As a "peer" package, the consumer of this library will need to perform `npm install kotlin` in their project, and they can pick any version they want.

twyatt:
@davidtaylor-juul is this working w/ the Gradle plugin now?

davidtaylor-juul:
> @davidtaylor-juul is this working w/ the Gradle plugin now?

Yes, this plugin had a bunch up updates and some new releases since I last tried it (in the last couple of days) and it's working now.

gradle/gitLatestTag.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+git tag | grep -E '^[0-9]' | sort -V | tail -1
```
twyatt, line 2: You shouldn't need a bash script for this.

https://github.com/JuulLabs/juul-app-android/blob/05589f86de2aad7453768d15a4e415fabc65f367/gradle/versions.gradle

koap/build.gradle.kts
```diff
@@ -58,3 +62,55 @@ kotlin {
...
+fun String.runCommand(workingDir: File = file("./")): String {
+    val parts = this.split("\\s".toRegex())
+    val proc = ProcessBuilder(*parts.toTypedArray())
+            .directory(workingDir)
+            .redirectOutput(ProcessBuilder.Redirect.PIPE)
+            .redirectError(ProcessBuilder.Redirect.PIPE)
+            .start()
+    proc.waitFor(1, TimeUnit.MINUTES)
+    return proc.inputStream.bufferedReader().readText().trim()
+}
```
twyatt, line 76: I'm hoping the `execute` function still works in Kotlin DSL. It would simplify this considerably.

https://github.com/JuulLabs/juul-app-android/blob/05589f86de2aad7453768d15a4e415fabc65f367/gradle/versions.gradle

koap/build.gradle.kts
```diff
@@ -58,3 +62,55 @@ kotlin {
...
+            }
+
+            dependencies {
+                npmPeer("kotlin", "*")
+            }
+
+            packageJson {
+                // TODO: set private = false when working
+                private = true
+                version = "${rootDir}/gradle/gitLatestTag.sh".runCommand()
```
twyatt, line 112: nitpick

```suggestion
                version = "$rootDir/gradle/gitLatestTag.sh".runCommand()
```

koap/build.gradle.kts
twyatt: This is a good note. Can you summarize and have it as a comment in the Gradle file, please?

koap/build.gradle.kts
```diff
@@ -1,9 +1,13 @@
+import lt.petuska.npm.publish.dsl.NpmAccess
```
twyatt, line 1: ultra nitpick

```suggestion
import lt.petuska.npm.publish.dsl.NpmAccess.RESTRICTED
```

Then below, `NpmAccess.RESTRICTED` can be replaced with `RESTRICTED`.

koap/build.gradle.kts
davidtaylor-juul: Unfortunately `execute` is a Groovy string enhancement: http://docs.groovy-lang.org/latest/html/groovy-jdk/java/lang/String.html#execute()

I can simplify it a bit by using `project.exec` and move it into a `buildSrc` file to get it out of the way.

gradle/gitLatestTag.sh
davidtaylor-juul: Yeah this was a hack as using `ProcessBuilder` does not allow for multiple piped commands so easily. I'll change it to use a sub shell and `project.exec` and delete this.

buildSrc/src/main/kotlin/GitInfo.kt
```diff
@@ -0,0 +1,14 @@
...
+    val byteOut = ByteArrayOutputStream()
+    project.exec {
+        commandLine = listOf( "sh", "-c", command)
+        standardOutput = byteOut
+    }
+    return byteOut.toString()
+}
+
+fun Project.gitLatestTag(): String =
+        runCommand("git tag | grep -E '^[0-9]' | sort -V | tail -1").trim()
```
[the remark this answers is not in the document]
davidtaylor-juul: Nice catch. I wanted a command that gets the highest numbered version tag, not just the most recent tag because I was originally including the `latest` publish config key in the npm package which has to always be semantically increasing. The naive pattern match here only works for the simple case, whoops!

We can live with the most recent tag instead (i.e. what `git describe` provides) now as I removed the `latest` tag from the npm package and it can publish more recent tags that refer to older versions just fine.

Git - several dozen ways to do everything and none of them are *quite* what you want. :-/

buildSrc/src/main/kotlin/GitInfo.kt
```diff
@@ -0,0 +1,14 @@
...
+    val byteOut = ByteArrayOutputStream()
+    project.exec {
+        commandLine = listOf( "sh", "-c", command)
+        standardOutput = byteOut
+    }
+    return byteOut.toString()
+}
+
+fun Project.gitLatestTag(): String =
+        runCommand("git describe --tags --abbrev=0").trim()
```
twyatt, line 14: nitpick: use 4-space indentation

```suggestion
    runCommand("git describe --tags --abbrev=0").trim()
```

twyatt (approved):

settings.gradle.kts
```diff
@@ -1,13 +1,18 @@
 pluginManagement {
     repositories {
         gradlePluginPortal()
+        maven {
+            url = uri("https://dl.bintray.com/mpetuska/lt.petuska")
+        }
```
twyatt, line 6: According to Martynas Petuška (author of `npm-publish`), [the plugin is now on Gradle plugin portal](https://kotlinlang.slack.com/archives/C0B8L3U69/p1603263212387900?thread_ts=1602689960.260900&cid=C0B8L3U69), so this should no longer be needed.

```suggestion
```

settings.gradle.kts
```diff
@@ -5,9 +5,11 @@ pluginManagement {
 
     resolutionStrategy {
         eachPlugin {
-            when {
-                requested.id.id == "binary-compatibility-validator" ->
+            when (requested.id.id) {
+                "binary-compatibility-validator" ->
                     useModule("org.jetbrains.kotlinx:binary-compatibility-validator:${requested.version}")
+                "lt.petuska.npm.publish" ->
```
mpetuska, line 11: No longer needed since it's on gradlePluginPortal

koap/build.gradle.kts
```diff
@@ -58,3 +61,41 @@ kotlin {
...
+        repository("github") {
+            registry = uri("https://npm.pkg.github.com")
+        }
+    }
+
+    publications {
+        publication(name) {
+            main = "$name.js"
+
+            files {
```
mpetuska, line 84: Can & should be removed for plugin v1.0.0

koap/build.gradle.kts
```diff
@@ -58,3 +61,41 @@ kotlin {
...
+npmPublishing {
+    organization = "juullabs"
+
+    repositories {
+        repository("github") {
+            registry = uri("https://npm.pkg.github.com")
+        }
+    }
+
+    publications {
```
mpetuska, line 80: Instead of creating a new publication, extend the preconfigured one 
```
//publication(name) {}
named("js") {
}
```

koap/build.gradle.kts
```diff
@@ -58,3 +61,41 @@ kotlin {
         }
     }
 }
+
+tasks.withType<NpmPackageAssembleTask> {
```
mpetuska, line 65: Not needed for plugin v1.0.0

koap/build.gradle.kts
```diff
@@ -58,3 +61,41 @@ kotlin {
...
+
+            dependencies {
+                // Peer dependencies: https://docs.npmjs.com/files/package.json#peerdependencies
+                // For any peer packages listed here, the consumer of this library will need to perform
+                // `npm install <pkg>` to pick up a version that matches the version pattern provided
+                npmPeer("kotlin", "^1.4")
+            }
+
+            packageJson {
+                version = gitMostRecentTag()
```
mpetuska, line 97: Can be declared on publication
```

**Comment by davidtaylor-juul** [comment]
Updated with v1.0.0 of the plugin which fixes the dependency packaging and allows removal of the webpack step.

File: .github/workflows/ci.yml [hunk 14]
```diff
@@ -74,3 +78,12 @@ jobs:
...
           -Psigning.secretKeyRingFile="$HOME/secring.gpg"
+
+      - name: Publish NPM
+        if: github.event_name == 'release' && startsWith(github.ref, 'refs/tags/')
+        run: >-
+          ./gradlew
+          $GRADLE_ARGS
+          publishJsNpmPublicationToGithub
+          -Pnpm.publish.authToken.github="${{ secrets.MOBILE_GITHUB_PACKAGES_PASSWORD }}"
+          -Pnpm.publish.version="${{ steps.get_version.outputs.VERSION }}"
```
twyatt, line 89: Rather than have a separate `step` ("Get the version" above), can we just use the variable directly here?

```suggestion
          -Pnpm.publish.version="${GITHUB_REF/refs\/tags\//}"
```

twyatt (commented):
Really love how minimal/simple the npm publication configuration ended up being. Will make it really easy to repeat on other libraries. 😄

koap/build.gradle.kts
```diff
@@ -58,3 +59,19 @@ kotlin {
...
+
+    repositories {
+        repository("github") {
+            registry = uri("https://npm.pkg.github.com")
+        }
+    }
+
+    publications {
+        named("js") {
+            version = project.findProperty("npm.publish.version").toString()
```
mpetuska, line 74: Instead of having to maintain this, why not just set the version at the top level? (Plugin uses project.version as default)

mpetuska: Or you can simply invoke the build from ci just passing -Pversion=xxx and it'll resolve the same with no comfig required in buildfile

mpetuska: That way the only thing you need to config is the repository

.github/workflows/ci.yml
twyatt: Possibly worth combining the two publish steps into one?

```
      - name: Publish
        if: startsWith(github.ref, 'refs/tags/')
        run: >-
          ./gradlew
          $GRADLE_ARGS
          uploadArchives
          publishJsNpmPublicationToGithub
          -PVERSION_NAME=${GITHUB_REF/refs\/tags\//}
          -Pversion="${GITHUB_REF/refs\/tags\//}"
          -Psigning.keyId="${{ secrets.SIGNING_KEY_ID }}"
          -Psigning.password="${{ secrets.SIGNING_PASSWORD }}"
          -Psigning.secretKeyRingFile="$HOME/secring.gpg"
```

I haven't checked, but the `com.vanniktech.maven.publish` plugin may work with `publish` task, which I believe would allow us to replace `uploadArchives` with `publish` and have it work for both publication to both Maven and npm.

koap/build.gradle.kts
twyatt: Thanks so much for all the help @mpetuska!

.github/workflows/ci.yml
twyatt: Alternatively, it should be simple enough to have `publish` task depend on both the Maven (`installArchives`) and npm (`publishJsNpmPublicationToGithub`).

- https://github.com/gradle/kotlin-dsl-samples/blob/master/samples/task-dependencies/build.gradle.kts
- https://docs.gradle.org/current/userguide/more_about_tasks.html

davidtaylor-juul: For sure we can use the pattern extraction directly without the extra step, have made that change. Will also use this to override the `project.version` variable as suggested.

As for combining the publishing tasks I'd recommend we keep it separate for the moment until we publish a live tag and verify that everything is operating standalone. Once that is confirmed we can go about optimizing the publishing tasks as a combined sequence in a separate PR.

koap/build.gradle.kts
davidtaylor-juul: FYI, the top level version is set as `unspecified` hence the requirement to inject it:
https://github.com/JuulLabs/koap/blob/dtaylor/apd-954_npm_pkg_on_main_petuska/gradle.properties#L2

Using `-Pversion=xxx` is the ideal solution, thanks for the tip. Will apply this change.

davidtaylor-juul: Please disregard my reference to the `gradle.properties` `VERSION_NAME` property - that has no bearing here.

Unfortunately the `-Pversion=xxx` option fails to override the `project.version` property, at least when it comes to the value that is emitted into the `package.json` file. It is unclear why this is but I suspect it is to do with the behavior of the Gradle Kotlin DSL configuration?

Will have to stick with `npm.publish.version` at least for now.

.github/workflows/ci.yml
mpetuska: FYI, publishToNpm tasks generated by the plugin already are added as dependencies to publish task :)

koap/build.gradle.kts
mpetuska: Apologies, looks like there's a bug in the plugin that's using rootProject.version instead on project.version as default. So yeah, stick with your solution until next release. You could set the version at publication.version instead of publication.packageJson.version to make it more concise

.github/workflows/ci.yml
twyatt: Quick follow up:

`npm-publish` conveniently makes sure npm tasks are configured as dependencies of `publish`, unfortunately if we want to keep our Maven and npm publications steps separate we'll have to explicitly call out the npm task.

In the future, we might try combining the publication (and use the official maven publish Gradle plugin) at which point we can use the `publish` task.

---

Demonstrating `publish` depending on our Maven publication tasks:

<details>
<summary>Gradle run</summary>

```
./gradlew -Pnpm.publish.authToken.github=foo publish --dry-run

:kotlinNodeJsSetup SKIPPED
:koap:jsPackageJson SKIPPED
:rootPackageJson SKIPPED
:kotlinNpmInstall SKIPPED
:koap:jsGenerateExternalsIntegrated SKIPPED
:koap:compileKotlinJs SKIPPED
:koap:jsProcessResources SKIPPED
:koap:assembleJsNpmPublication SKIPPED
:koap:publishJsNpmPublicationToGithub SKIPPED
:koap:jsMainClasses SKIPPED
:koap:jsPublicPackageJson SKIPPED
:koap:jsJar SKIPPED
:koap:generateMetadataFileForJsPublication SKIPPED
:koap:generatePomFileForJsPublication SKIPPED
:koap:compileJava SKIPPED
:koap:processResources SKIPPED
:koap:classes SKIPPED
:koap:javadoc SKIPPED
:koap:javadocsJar SKIPPED
:koap:jsSourcesJar SKIPPED
:koap:signJsPublication SKIPPED
:koap:publishJsPublicationToLocalRepository SKIPPED
:koap:publishJsPublicationToMavenRepository SKIPPED
:koap:compileKotlinJvm SKIPPED
:koap:jvmProcessResources SKIPPED
:koap:jvmMainClasses SKIPPED
:koap:jvmJar SKIPPED
:koap:generateMetadataFileForJvmPublication SKIPPED
:koap:generatePomFileForJvmPublication SKIPPED
:koap:jvmSourcesJar SKIPPED
:koap:signJvmPublication SKIPPED
:koap:publishJvmPublicationToLocalRepository SKIPPED
:koap:publishJvmPublicationToMavenRepository SKIPPED
:koap:emptySourcesJar SKIPPED
:koap:generateMetadataFileForKotlinMultiplatformPublication SKIPPED
:koap:generatePomFileForKotlinMultiplatformPublication SKIPPED
:koap:signKotlinMultiplatformPublication SKIPPED
:koap:publishKotlinMultiplatformPublicationToLocalRepository SKIPPED
:koap:publishKotlinMultiplatformPublicationToMavenRepository SKIPPED
:koap:compileKotlinMetadata SKIPPED
:koap:metadataMainClasses SKIPPED
:koap:metadataJar SKIPPED
:koap:generateMetadataFileForMetadataPublication SKIPPED
:koap:generatePomFileForMetadataPublication SKIPPED
:koap:metadataSourcesJar SKIPPED
:koap:signMetadataPublication SKIPPED
:koap:publishMetadataPublicationToLocalRepository SKIPPED
:koap:publishMetadataPublicationToMavenRepository SKIPPED
:koap:publish SKIPPED
```
</details>

koap/build.gradle.kts
davidtaylor-juul: > You could set the version at publication.version instead of publication.packageJson.version to make it more concise

I tried this and it didn't work, but that's ok we can wait for the next patch and delete the entire section anyhow.

mpetuska: I see. Thanks for letting me know. Something to look into and fix in the next release

twyatt:
@davidtaylor-juul just fyi, npm-publish [1.0.1](https://gitlab.com/lt.petuska/npm-publish/-/releases/v1.0.1) has been released.

koap/build.gradle.kts
```diff
@@ -32,7 +33,7 @@ kotlin {
         val commonMain by getting {
             dependencies {
                 api(kotlin("stdlib"))
-                implementation("com.squareup.okio:okio-multiplatform:2.6.0")
+                implementation("com.squareup.okio:okio-multiplatform:2.9.0")
```
QuantumRand, line 36: Change unrelated?

QuantumRand (approved):
Amazing how concise this is!

koap/build.gradle.kts
twyatt: We weren't sure if some of the npm dependency packaging issues were plugin related or specific to our dependencies. The version bump was hoping that older okio version weren't properly configured for multiplatform. I don't believe that ended up being the case; so this version bump was unnecessary (for the purposes of this PR).

davidtaylor-juul: So should we undo this?

twyatt: It is outside the scope of this PR. For me it's a `nit`, I'm fine with leaving it in as we should upgrade dependencies. If it's not too much trouble, then removing it from this PR is better.

davidtaylor-juul: Make sense. Given this repo is very much in a pre-beta state I'll go ahead and leave it in.

PR #35, opened 2020-10-27
APD-957: Add JsExport Annotations

davidtaylor-juul:
Adds `@JsExport` annotations to explicitly state the Javascript API.

From https://kotlinlang.org/docs/reference/js-to-kotlin-interop.html#jsexport-annotation
> The @JsExport annotation is available in the current default compiler backend and the new IR compiler backend. If you are targeting the IR compiler backend, you must use the @JsExport annotation to make your functions visible from Kotlin in the first place.

In order to emit typescript type declarations we need to be using the IR compiler which requires these annotations, so this is one step towards that.

More importantly, these annotations are also understood by the legacy compiler and using them forces the code author to explicitly note what is and isn't available to Javascript.

Where feasible `@file:JsExport` has been used. Where a file contains public declarations that are not supported, the declarations that are supported have been annotated individually.

One messy refactor is replacement of the nested `enum` classes on the `Message` object as `enum` is not supported and using it precludes export of the `Message` class entirely.

There is more refactoring required in order to have the entire API exported, or removal/relocation of Kotlin specific language features that are nice-to-haves but can be done without. That step is best left for a separate PR.

### DIFF
--- a/koap/src/commonMain/kotlin/Decoder.kt
+++ b/koap/src/commonMain/kotlin/Decoder.kt
@@ -47,6 +47,8 @@ import com.juul.koap.Message.Udp.Type.Acknowledgement
 import com.juul.koap.Message.Udp.Type.Confirmable
 import com.juul.koap.Message.Udp.Type.NonConfirmable
 import com.juul.koap.Message.Udp.Type.Reset
+import kotlin.js.JsExport
+import kotlin.js.JsName
 import okio.BufferedSource
 
 /**
@@ -97,6 +99,7 @@ inline fun <reified T : Message> ByteArray.decode(): T =
  * +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
  * ```
  */
+@JsExport
 fun ByteArray.decodeUdp(): Message.Udp {
     val header = decodeUdpHeader()
     return decode(header, offset = header.size)
@@ -120,6 +123,7 @@ fun ByteArray.decodeUdp(): Message.Udp {
  * +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
  * ```
  */
+@JsExport
 fun ByteArray.decodeTcp(): Message.Tcp {
     val header = decodeTcpHeader()
     return decode(header, offset = header.size)
@@ -146,6 +150,8 @@ fun ByteArray.decodeTcp(): Message.Tcp {
  * val message = encoded.decode(content, offset = 0)
  * ```
  */
+@JsExport
+@JsName("decodeWithUdpHeader")
 fun ByteArray.decode(
     header: Header.Udp,
     offset: Int = header.size
@@ -172,6 +178,8 @@ fun ByteArray.decode(
  * val message = content.decode(content, offset = 0)
  * ```
  */
+@JsExport
+@JsName("decodeWithTcpHeader")
 fun ByteArray.decode(
     header: Header.Tcp,
     offset: Int = header.size
@@ -229,6 +237,7 @@ private fun ByteArray.decodeContent(
  * +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
  * ```
  */
+@JsExport
 fun ByteArray.decodeUdpHeader(): Header.Udp = withReader {
     // |7 6 5 4 3 2 1 0|
     // +-+-+-+-+-+-+-+-+
@@ -279,6 +288,7 @@ fun ByteArray.decodeUdpHeader(): Header.Udp = withReader {
  * +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
  * ```
  */
+@JsExport
 fun ByteArray.decodeTcpHeader(): Header.Tcp = withReader {
     // |7 6 5 4 3 2 1 0|
     // +-+-+-+-+-+-+-+-+
--- a/koap/src/commonMain/kotlin/Decoder.kt
+++ b/koap/src/commonMain/kotlin/Decoder.kt
@@ -47,9 +47,9 @@ import com.juul.koap.Message.Udp.Type.Acknowledgement
 import com.juul.koap.Message.Udp.Type.Confirmable
 import com.juul.koap.Message.Udp.Type.NonConfirmable
 import com.juul.koap.Message.Udp.Type.Reset
+import okio.BufferedSource
 import kotlin.js.JsExport
 import kotlin.js.JsName
-import okio.BufferedSource
 
 /**
  * Decodes [ByteArray] receiver to a [Message].
--- a/koap/src/commonMain/kotlin/Encoder.kt
+++ b/koap/src/commonMain/kotlin/Encoder.kt
@@ -13,10 +13,9 @@ import com.juul.koap.Message.Udp.Type.Acknowledgement
 import com.juul.koap.Message.Udp.Type.Confirmable
 import com.juul.koap.Message.Udp.Type.NonConfirmable
 import com.juul.koap.Message.Udp.Type.Reset
-import kotlin.js.JsExport
-import kotlin.js.JsName
 import okio.Buffer
 import okio.BufferedSink
+import kotlin.js.JsExport
 
 internal const val UINT32_MAX_EXTENDED_LENGTH = UINT_MAX_VALUE + 65805L
 
--- a/koap/src/commonMain/kotlin/Encoder.kt
+++ b/koap/src/commonMain/kotlin/Encoder.kt
@@ -1,3 +1,4 @@
+@file:JsExport
 package com.juul.koap
 
 import com.juul.koap.Message.Option
@@ -12,6 +13,8 @@ import com.juul.koap.Message.Udp.Type.Acknowledgement
 import com.juul.koap.Message.Udp.Type.Confirmable
 import com.juul.koap.Message.Udp.Type.NonConfirmable
 import com.juul.koap.Message.Udp.Type.Reset
+import kotlin.js.JsExport
+import kotlin.js.JsName
 import okio.Buffer
 import okio.BufferedSink
 
--- a/koap/src/commonMain/kotlin/Encoder.kt
+++ b/koap/src/commonMain/kotlin/Encoder.kt
@@ -1,4 +1,5 @@
 @file:JsExport
+
 package com.juul.koap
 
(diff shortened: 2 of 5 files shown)


koap/src/commonMain/kotlin/Decoder.kt
```diff
@@ -97,6 +99,7 @@ inline fun <reified T : Message> ByteArray.decode(): T =
  * +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
  * ```
  */
+@JsExport
```
twyatt, line 102: Particular reason for not doing file-wide export?

koap/src/commonMain/kotlin/Message.kt
```diff
@@ -200,6 +201,7 @@ sealed class Message {
         /** RFC 7252 5.10.4. Accept */
         data class Accept(val format: Long) : Option() {
 
+            @JsName("constructorWithContentFormat")
```
twyatt, line 204: With this annotation, how is this accessed from JS?

_I feel like there is probably a better name, as you generally don't have "constructor" in function names._

koap/src/commonMain/kotlin/Message.kt
```diff
@@ -291,17 +293,21 @@ sealed class Message {
              * - `0` (register) adds the entry to the list, if not present;
              * - `1` (deregister) removes the entry from the list, if present.
              */
-            enum class Registration { Register, Deregister }
```
twyatt, line 294: I'm a little confused by this. I thought `enum`s from Kotlin would work fine in JS?
The `Enum` class itself is marked as supporting JS platform.

https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-enum/

koap/src/commonMain/kotlin/Decoder.kt
davidtaylor-juul: https://github.com/JuulLabs/koap/blob/dtaylor/apd-957_typescript_types/koap/src/commonMain/kotlin/Decoder.kt#L78

> Decoder.kt: (79, 1): Declaration of such kind (inline function with reified type parameters) cant be exported to JS

There are various techniques we can employ to deal with this but I wanted to leave that for a separate PR.

koap/src/commonMain/kotlin/Message.kt
davidtaylor-juul: I was hoping to get some suggestions for the naming or a different strategy here TBH. The `JsName` here will map to a class function at the immediate parent scope, so in this case it would be:
```js
import koapModule from "..."
let acceptOption = koapModule.Message.Option.constructorWithContentFormat(0)
```

This is clearly terrible 🤪 

We should probably find a convention for these, perhaps `Message.Option.newAcceptOptionWithContentFormat`?

koap/src/commonMain/kotlin/Message.kt
davidtaylor-juul: This is the error:
> Message.kt: (296, 24): Declaration of such kind (enum class) cant be exported to JS

¯\\_(ツ)_/¯ 

I think it's objecting due to it being nested. I took the minimal route of declaring it as a `sealed class` but an alternative would probably be to move all enums to the root level. Again, probably best left to a separate PR to do that kind of refactoring.

koap/src/commonMain/kotlin/Message.kt
twyatt: Aw, I see the dilemma now.

Being that JS isn't type-safe we can't use type overloading. 😢 

If we export the primary constructor only (`data class Accept(val format: Long)`) will the secondary constructor (`constructor(format: ContentFormat)`) be hidden in JS?

I ask, because `ContentFormat` itself is just a wrapper around the `format` Long value, so if we can just hide the secondary constructor in JS then JS users could:

```typescript
import koapModule from "..."
let acceptOption = koapModule.Message.Option.Accept(0)

// or
let contentFormat = koapModule.Message.Option.ContentFormat(0)
koapModule.Message.Option.Accept(contentFormat.format)
```

---

If we can't only have the primary constructor visible, then my next idea would be to drop the secondary constructor (in common) then add an extension function in `jvmMain`:

```kotlin
fun Message.Option.Accept(format: ContentFormat) = Accept(format.format)
```

koap/src/commonMain/kotlin/Message.kt
twyatt: Oh, looks like `enum`s just can't be exported _yet_. So I don't think it's a nested issue.

https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.js/-js-export/

---

`sealed class` is a good alternative. 👍 
My original hesitation was I was hoping we could get the JS target going w/o breaking API changes; looks like this will just have to be a major version bump (which is fine).

koap/src/commonMain/kotlin/Header.kt
```diff
@@ -1,7 +1,9 @@
+@file:JsExport
```
twyatt, line 1: ultra ultra ultra nitpick: newline after

```suggestion
@file:JsExport

```

koap/src/commonMain/kotlin/Encoder.kt
```diff
@@ -1,3 +1,4 @@
+@file:JsExport
```
twyatt, line 1: ultra ultra ultra nitpick

```suggestion
@file:JsExport

```

koap/src/commonMain/kotlin/Message.kt
```diff
@@ -291,17 +293,21 @@ sealed class Message {
...
              *
              * @see Registration
              */
+            @JsName("constructorWithRegistration")
             constructor(action: Registration) : this(
                 when (action) {
-                    Register -> 0L
-                    Deregister -> 1L
+                    Message.Option.Observe.Registration.Register -> 0L
+                    Message.Option.Observe.Registration.Deregister -> 1L
```
twyatt, line 310: Add appropriate imports so can be written as:

```suggestion
                    Register -> 0L
                    Deregister -> 1L
```

koap/src/commonMain/kotlin/Message.kt
```diff
@@ -291,17 +293,21 @@ sealed class Message {
...
+                object Register: Registration()
+                object Deregister: Registration()
+            }
 
             /**
              * Constructs an [Observe] to be included in a GET request.
              *
              * @see Registration
              */
+            @JsName("constructorWithRegistration")
```
twyatt, line 306: Rather than expose this constructor, can we take a similar approach as https://github.com/JuulLabs/koap/pull/35/files#r513669441?

koap/src/commonMain/kotlin/Message.kt
```diff
@@ -389,11 +395,11 @@ sealed class Message {
         override val payload: ByteArray
     ) : Message() {
 
-        enum class Type {
-            Confirmable,
-            NonConfirmable,
-            Acknowledgement,
-            Reset,
+        sealed class Type() {
```
twyatt, line 398:
```suggestion
        sealed class Type {
```

koap/src/commonMain/kotlin/Message.kt
davidtaylor-juul: > If we export the primary constructor only (data class Accept(val format: Long)) will the secondary constructor (constructor(format: ContentFormat)) be hidden in JS?

Both constructors are exported due to the enclosing class being annotated for export. It doesn't seem to be possible to hide a specific nested symbol (i.e. override the enclosing export).

The primary constructor is bound to the object initializer, so in javascript they would look like this:
```js
import koapModule from "..."
// primary (requires `new` keyword)
let acceptOption1 = new koapModule.Message.Option.Accept(0)
// secondary (class method)
let acceptOption2 = koapModule.Message.Option.newAcceptOptionWithContentFormat(
    new koapModule.Message.Option.ContentFormat(0)
)
```

> If we can't only have the primary constructor visible, then my next idea would be to drop the secondary constructor (in common) then add an extension function in `jvmMain`

Perhaps. But then we should consider if it is needed at all rather than introduce asymmetric APIs. If keeping it has merit, should we refactor things such that js consumers are provided with a nicer/safer interface in its place? ¯\\_(ツ)_/¯ 

Again, questions that lead to rethinking and refactoring which I'm deliberately avoiding in this PR. :smile: Answering coherently requires studying the use-cases as there are likely knock-on effects. The goal here is to enable the feature so that the compiler can insert its opinions during future adjustments to the code.

koap/src/commonMain/kotlin/Header.kt
davidtaylor-juul: I originally had a newline but the IDE complained 🤪

koap/src/commonMain/kotlin/Message.kt
davidtaylor-juul: Surprised the linter didn't pick up on this one.

koap/src/commonMain/kotlin/Header.kt
twyatt: Weird. No newline is fine if it's absence is needed to quiet the IDE.

koap/src/commonMain/kotlin/Message.kt
twyatt: I think the standard set of lint rules is pretty conservative, i.e. there is a much larger set of "experimental" lint rules we can enable (maybe we should), and I'd hope a lint rule is in there that would catch this. 🤷

koap/src/commonMain/kotlin/Message.kt
cedrickcooke: Considering the goal of not doing a major refactor, I propose we give it a name like we would from a companion object. In this case, `fromContentFormat`.

koap/src/commonMain/kotlin/Message.kt
cedrickcooke: Again, I like treating it as though it was a function on the companion object for JS. In this case, `fromRegistration`, and calling it as `Observe.fromRegistration(registration)`

cedrickcooke (approved):
Looks good to me, apart from the bikeshedding on the names.

koap/src/commonMain/kotlin/Header.kt
davidtaylor-juul: Super weird. And since merging latest `main` it seems happy with a newline now.

koap/src/commonMain/kotlin/Message.kt
davidtaylor-juul: Concur, will go with this naming approach.

davidtaylor-juul:
Q: what is the process with `apiCheck` failing, is there an official step required to acknowledge the breaking changes before merging?

twyatt (approved):

twyatt:
> Q: what is the process with `apiCheck` failing, is there an official step required to acknowledge the breaking changes before merging?

```
./gradlew apiDump
```

PR #38, opened 2020-11-06
Validation tests for JavaScript target public API

Phoenix7351:
Adding tests to ensure that the generated JS package outputs functions that are actually usable from the JS environment.

Incorporating these tests into CI.


koap/js_tests/decodeTcp.test.js
```diff
@@ -0,0 +1,39 @@
+let koapModule = require('../build/distributions/koap')
```
Phoenix7351, line 1: I haven't been able to get "import" properly working.

koap/js_tests/decodeTcp.test.js
```diff
@@ -0,0 +1,39 @@
...
+const koap = koapModule.com.juul.koap
+
+let kotlinModule = require('../build/publications/npm/js/node_modules/kotlin')
+const kotlin = kotlinModule
+
+test('TCP empty payload :: correctly decoded', () => {
```
Phoenix7351, line 7: The testing format I'm going with atm is:

Function being tested is notated by the file name. In this case, decodeTcp
'TCP empty payload' is denoting the special input conditions of the test
'correctly decoded' is denoting what results are being tested for.

koap/js_tests/decodeTcp.test.js
```diff
@@ -0,0 +1,39 @@
...
+  
+    let encodedMessage = new Uint8Array([0xA1, 0x02, 0x42, 0xB4, 0x69, 0x6E, 0x66, 0x6F, 0x44, 0x62, 0x61, 0x74, 0x74])
+
+    let decodedMessage = koap.decodeTcp(encodedMessage)
+
+    console.log(decodedMessage)
+
+    // Better to not convert? Introduces extra point of failure this way
+    let decodedPayload = byteArrayToHexString(decodedMessage.payload)
```
Phoenix7351, line 16: I think I might stop doing this. While byteArrayToHexString has been useful for me while looking at the results, as I commented, it's adding a potential point of failure. I.e. if that function gets a bug, tests will start failing incorrectly.

koap/js_tests/package.json
```diff
@@ -0,0 +1,30 @@
+{
```
Phoenix7351, line 1: I messed around a lot in this file trying to get importing working. There's some stuff I probably don't need.

koap/build.gradle.kts
```diff
@@ -11,7 +11,14 @@ apply(from = rootProject.file("gradle/jacoco.gradle.kts"))
 
 kotlin {
     jvm()
-    js().browser()
+    js {
+        browser {
+            webpackTask {
+                output.libraryTarget = "commonjs2"
```
twyatt, line 17: Why the switch to `commonjs2`?

I was under the impression that UMD (which I think we were configured with by default) is compatible with CommonJS?

koap/js_tests/decodeTcp.test.js
twyatt: Does JS have assertions that work with `UInt8Array`? If so, then you're right, we should probably stick with it being `UInt8Array` (a.k.a. `ByteArray`) rather than converting it to a String.

twyatt (commented):
Very cool. Great job!

General feedback:
- Do we need `koap/js_tests/package-lock.json`?
    - It's crazy big and looks like it would result in a **lot** of version control churn
- We could probably get away with fewer tests, since a lot of functionality will be tested by Kotlin unit tests
    - I'd imagine we only need a couple of tests just to validate the API surface works as expected from JS, right?

koap/js_tests/encode.test.js
```diff
@@ -0,0 +1,110 @@
...
+  let encoded = koap.encode(message)
+
+  // Better to not convert? Introduces extra point of failure this way
+  let hexEncoded = byteArrayToHexString(encoded)
+  console.log(hexEncoded)
+  
+  expect(encoded).toEqual(new Int8Array([0xD1, 0x04, 0x02, 0x42, 0xB4, 0x69, 0x6E, 0x66, 0x6F, 0x44, 0x62, 0x61, 0x74, 0x74, 0xFF, 0x42, 0x61, 0x74, 0x6D, 0x61, 0x6E]));
+});
+
+test('UDP GET empty payload :: correct encoding', () => {
```
earthtosid, line 82: Test comment

koap/js_tests/decodeTcp.test.js
Phoenix7351: Yeah, pretty sure jest can check equivalence between two arrays. Gonna change this out.

Phoenix7351:
> Very cool. Great job!
> 
> General feedback:
> 
> * Do we need `koap/js_tests/package-lock.json`?
>   
>   * It's crazy big and looks like it would result in a **lot** of version control churn
> * We could probably get away with fewer tests, since a lot of functionality will be tested by Kotlin unit tests
>   
>   * I'd imagine we only need a couple of tests just to validate the API surface works as expected from JS, right?

1. Probably not? Might need someone with more npm experience to answer that.

2. Mmm. Not sure I agree. This effort is really testing a new product, which is why we're taking this approach anyway--an undercurrent of fear that something will go wrong in the transpilation and/or programmer error as a result of that process. Sure, you've got tests covering the Kotlin end, but they'll do little to protect the validity of this product. The way I'm approaching it, then, is just regarding it as its own thing that needs to have all the angles covered.

koap/js_tests/encode.test.js
Phoenix7351: Test response.

koap/build.gradle.kts
Phoenix7351: It was something that @davidtaylor-juul had me do. I can't precisely remember the reason.

twyatt:
> an undercurrent of fear that something will go wrong in the transpilation

The Kotlin unit tests **are** transpiled when testing the JS target. So we already do test the JS transpiled code with our Kotlin unit tests.

The problem arrises with name mangling that occurs and can make the public API surface malformed. My understanding is that the tests written directly in JS (not transpiled) are to validate that the API surface has the shape that was expected.

koap/js_tests/decodeTcp.test.js
davidtaylor-juul: Jest has a deep equality check for this purpose e.g.
```js
let expected = new Uint8Array([...])
expect(decodedMessage.payload).toStrictEqual(expected)
```

davidtaylor-juul:
> The problem arrises with name mangling that occurs and can make the public API surface malformed. My understanding is that the tests written directly in JS (not transpiled) are to validate that the API surface has the shape that was expected.

^this
These tests were inspired by the initial published artifacts being unusable from JS. There are certain Kotlin language features (e.g. enums, overloading) that do not provide JS interop by default (yet). The intention is to validate that the published artifacts bridge to JS with a usable API.

davidtaylor-juul:
> Do we need koap/js_tests/package-lock.json?

Conventional wisdom is to always commit the lock file for JS projects to ensure dependency matching for any npm packages that are loosely versioned in `package.json`. For unit tests in this situation I'd say that rule should be thrown out the window with tremendous enthusiasm.

Phoenix7351:
> > The problem arrises with name mangling that occurs and can make the public API surface malformed. My understanding is that the tests written directly in JS (not transpiled) are to validate that the API surface has the shape that was expected.
> 
> ^this
> These tests were inspired by the initial published artifacts being unusable from JS. There are certain Kotlin language features (e.g. enums, overloading) that do not provide JS interop by default (yet). The intention is to validate that the published artifacts bridge to JS with a usable API.

But in that case I still don't see the need to _not_ test thoroughly. As:

"There are certain Kotlin language features (e.g. enums, overloading) that do not provide JS interop by default (yet)"

We would want to be catching any of these sorts of things. And

"The Kotlin unit tests are transpiled when testing the JS target."

Maybe I'm not fully understanding, but couldn't those unit tests--themselves--be a point of failure since they're being transpiled as well?

Besides all this, what does having less coverage gain us? A little bit of time?

davidtaylor-juul:
>  I'd imagine we only need a couple of tests just to validate the API surface works as expected from JS, right?

Yes. Primarily to validate that the published API has clean JS interop is the goal. This *may* become unnecessary if every file with `public` symbols can cleanly compile with `@file:JsExport` added.

The secondary function is a more subjective analysis. It is very easy to write idiomatic Kotlin that results in supremely ugly JS API which hinders adoption. This is probably better achieved via a demo application though so the use cases can been seen in-situ.

davidtaylor-juul:
What do people think about a linter rule that enforces `@file:JsExport` for any file that has public symbols? Any unsupported language features would fail compilation and force the developer to tweak the API such that is it 100% javascript compatible.

twyatt:
> What do people think about a linter rule that enforces `@file:JsExport` for any file that has public symbols? Any unsupported language features would fail compilation and force the developer to tweak the API such that it is 100% javascript compatible.

Sounds good to me. 👍

koap/js_tests/helpers.js
```diff
@@ -0,0 +1,7 @@
+global.byteToHexString = (byte) => {
```
earthtosid, line 1: Global modifications may bleed into the rest of the application. These can be defined as functions that can be imported into your tests.

earthtosid (commented):
I'll do some digging to figure out the export situation. You may need `webpack` to bundle the tests and `babel` to output to them.

koap/js_tests/helpers.js
Phoenix7351: Well, I might not need these helpers at all.

koap/build.gradle.kts
```diff
@@ -11,7 +11,7 @@ apply(from = rootProject.file("gradle/jacoco.gradle.kts"))
 
 kotlin {
     jvm()
-    js().browser()
```
Phoenix7351, line 14: Oh, was this important?

koap/js_tests/decoder.test.ts
```diff
@@ -0,0 +1,105 @@
...
+
+    let encodedMessage = new Uint8Array([0xD1, 0x04, 0x02, 0x42, 0xB4, 0x69, 0x6E, 0x66, 0x6F, 0x44, 0x62, 0x61, 0x74, 0x74, 0xFF, 0x42, 0x61, 0x74, 0x6D, 0x61, 0x6E])
+
+    let decodedHeader = koap.decodeTcpHeader(encodedMessage)
+
+    expect(decodedHeader.size).toEqual(4)
+    expect(decodedHeader.length).toEqual(new kotlin.Long(17))
+    expect(decodedHeader.code).toEqual(koap.Message.Code.Method.POST)
+    expect(decodedHeader.token).toEqual(new kotlin.Long(66))
+});
```
Phoenix7351, line 105: Is it best practice to have a newline at the end of typescript files?

twyatt: Yes, best practice is to have newline at the end of any non-binary file.

koap/js_tests/package.json
```diff
@@ -0,0 +1,38 @@
...
+  "author": "Justin Tiffner",
+  "license": "ISC",
+  "devDependencies": {
+    "@babel/core": "^7.12.3",
+    "@babel/preset-env": "^7.12.1",
+    "@babel/preset-typescript": "^7.12.1",
+    "@types/jest": "^26.0.15",
+    "babel-jest": "^26.6.2",
+    "babel-loader": "^8.2.1",
+    "jest": "25.5.4",
```
Phoenix7351, line 26: I was getting warnings from ts-jest that it hadn't been tested with jest 26+ and to move back to <26.

Same thing with typescript 4.

koap/build.gradle.kts
twyatt: Yup. In future versions either `node()` or `browser()` will be required. Right now it's just a warning if you omit it.

Phoenix7351: Putting it back. ^_^

koap/js_tests/package.json
Phoenix7351: Everything in here should be needed now.

koap/js_tests/prepare-testing.sh
```diff
@@ -0,0 +1,13 @@
...
+
+./gradlew -Pnpm.publish.repository.github.authToken=fakingit -Pnpm.publish.version=0.0.3-test24 assemble
+
+cd -
+
+npm uninstall @juullabs/koap
+
+npm install file://../build/publications/npm/js
+
+npm run test
```
twyatt, line 13: This should become a Gradle task (maybe `tsTest` or similar name) instead of a bash script, so that `check` task can depend on it — and then it will be automatically executed for the current CI setup and under the standard `check` task run.

twyatt (commented):
Fantastic work!

Just some minor comments:
- Perhaps the `koap/js_tests` directory should be renamed to `tsTests` (since it's technically Typescript)
- Migrate from bash script to Gradle task

Phoenix7351:
> Fantastic work!
> 
> Just some minor comments:
> 
> * Perhaps the `koap/js_tests` directory should be renamed to `tsTests` (since it's technically Typescript)
> * Migrate from bash script to Gradle task

For the js_tests dir name change, kinda depends on how you look at it. My intention with the "js" was that I'm testing the js output--not that these are--in and of themselves--js tests. So maybe the best name is something like, "koap-js-tests?"

twyatt:
The `koap-` prefix on `koap-js-tests` feels redundant.
Technically you are testing from the perspective of Typescript, so `tsTests` feels appropriate to me; but also fine w/ `js` prefix if you'd prefer.
I suggested camel case as it's standard convention for JVM and Kotlin projects, so is more familiar; but being that this isn't being processed by the Kotlin toolchain, I'm fine with alternate naming conventions.

Phoenix7351:
> The `koap-` prefix on `koap-js-tests` feels redundant.
> Technically you are testing from the perspective of Typescript, so `tsTests` feels appropriate to me; but also fine w/ `js` prefix if you'd prefer.
> I suggested camel case as it's standard convention for JVM and Kotlin projects, so is more familiar; but being that this isn't being processed by the Kotlin toolchain, I'm fine with alternate naming conventions.

I see what you're saying, but I named it based on what's being tested instead of what the tests are written in--which I think would be more descriptive since tsTests doesn't offer up the purpose of the tests? That's how I'm looking at it, but I can change it if you like.

twyatt:
> I see what you're saying, but I named it based on what's being tested instead of what the tests are written in--which I think would be more descriptive since tsTests doesn't offer up the purpose of the tests? That's how I'm looking at it, but I can change it if you like.

Good call. With that in mind, maybe we should call it `apiTests` or `api-tests`. Then we're actually describing what the intent of the test suite is (to test/validate the API surface).

koap/build.gradle.kts
```diff
@@ -53,3 +53,15 @@ npmPublishing {
...
 }
+
+tasks.register("run-js-validation-testing") {
+
+    description = "Runs the package for testing"
+    group = "Verification"
+    dependsOn("assemble")
+
+    doLast {
+        var response = runCommand("../gradlew -Pnpm.publish.repository.github.authToken=fakingit -Pnpm.publish.version=0.0.3-test24 assemble && cd koap_js_tests && npm uninstall @juullabs/koap && npm install file://../build/publications/npm/js && npm run test")
```
twyatt, line 64: You shouldn't need the `../gradlew -Pnpm.publish.repository.github.authToken=fakingit -Pnpm.publish.version=0.0.3-test24 assemble` portion, as your Gradle task is already depending on `assemble`, so it will execute before this task.

twyatt, line 64: Does `npm install` not overwrite an existing install? Do you have to `uninstall` first every time?

Phoenix7351: I need a means to provide the option to set the published version to something. Without it, npm won't install it as it's version is "unspecified."

Phoenix7351: Honestly not sure. I'll try it out. @earthtosid @davidtaylor-juul Are there any reasons that we need to uninstall? I can't remember if that came up specifically on Friday.

twyatt: For just running the tests, `unspecified` seems fine for a version. Unless that breaks other things?

twyatt: Is `assemble` packaging the npm as a single file?
If so, couldn't we run the test without having to deal the with the install?
https://stackoverflow.com/a/28739855

earthtosid:
Wonder if we should setup `ESLint` for all the typescript stuff. [Here's a great article](https://khalilstemmler.com/blogs/typescript/eslint-for-typescript/) to able to do so. You can add `lint` and `fix` npm scripts in your package that could theoretically run before each commit. Or before a build in the CI/CD pipeline.

koap/build.gradle.kts
Phoenix7351: Without injecting a version, the npm install command fails with:

`> Task :koap:run-js-validation-testing FAILED
npm ERR! Invalid version: "unspecified"

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/justin.tiffner/.npm/_logs/2020-11-17T18_08_03_817Z-debug.log`

davidtaylor-juul: The npm package install will fail without a valid version number in the `package.json` file, so the command line version specifier is absolutely required. 

However as this depends on `assemble` there is no need to invoke it explicitly - it should be fine to just invoke this test task and provide the `-Pnpm.publish.version=<whatever>` argument to Gradle so that the npm assemble task has a version when it executes.

Phoenix7351: Running without the uninstall works fine, but I'll leave that decision to more experienced minds.

As for the assemble question: which npm/script are you referring to? The js api output or the test code?

davidtaylor-juul: The npm system has aggressive caching and will not overwrite if the version number has not changed, hence the advice to uninstall/reinstall. There are other ways to force it such as removing the package lock file that may or may not be faster if performance is a concern (it probably isn't).

Regarding the change here to use `npm install` I can explain. The previous iteration of these tests was using the Webpack artifacts (a single js file), but this was a hack due to the inability to load the non-webpack artifacts which are the ones that actually ship in the npm package (a structured npm multi-file/folder package). There is a very strong preference to test the module that actually ships. This is particularly important now that we are using npm shrink wrap hacks to access the bundled dependencies as it may affect visibility of those dependencies.

To do so, it is necessary to import the module that is emitted via the npm publish task. Due to security restrictions with the node runtime it is not possible to simply import the code from the build artifact folder (you cannot escape to the parent folder via `../`). So it then becomes necessary to perform the npm install using a `file://` url in order to be able to import the module into the javascript tests.

Phoenix7351: Well, that's what I was getting at with my first response. I'm assuming I can't retroactively provide this argument to assemble here in the test task. This task dependsOn assemble, so won't be executed until after, right? 

Would "doFirst" execute before the dependsOn?

Phoenix7351:
> Wonder if we should setup `ESLint` for all the typescript stuff. [Here's a great article](https://khalilstemmler.com/blogs/typescript/eslint-for-typescript/) to able to do so. You can add `lint` and `fix` npm scripts in your package that could theoretically run before each commit. Or before a build in the CI/CD pipeline.

I've got it added, but what rule set do we want to use?

koap/build.gradle.kts
twyatt: > As for the assemble question: which npm/script are you referring to? The js api output or the test code?

David answered my question.

So, it sounds like we should drop a lot of the shell commands and just depend on the `packJsNpmPublication` Gradle task.

Then you could have something like:

```kotlin
tasks.register("publishToLocalNpm") {
    dependsOn("packJsNpmPublication")
    runCommand("npm install -force file://../build/publications/npm/js")
}

tasks.register("apiTest") {
    dependsOn("publishToLocalNpm")
    runCommand("npm run tests")
}
```

Phoenix7351: Uhh. packJsNpmPublication--itself--fails because of "unspecified version."

`
Justins-MBP:koap justin.tiffner$ ./gradlew packJsNpmPublication

> Configure project :koap
Kotlin Multiplatform Projects are an Alpha feature. See: https://kotlinlang.org/docs/reference/evolution/components-stability.html. To hide this message, add 'kotlin.mpp.stability.nowarn=true' to the Gradle properties.

NPM Repository [github] is invalid. Skipping...

> Task :koap:packJsNpmPublication FAILED
npm ERR! Invalid version: "unspecified"

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/justin.tiffner/.npm/_logs/2020-11-18T00_46_26_101Z-debug.log
`

davidtaylor-juul: > Would "doFirst" execute before the dependsOn?

No. The `doFirst/Last` closures are executed when the task runs. Anything a task `dependsOn` runs before the task. https://docs.gradle.org/current/dsl/org.gradle.api.DefaultTask.html#N1DB72

.github/workflows/ci.yml
```diff
@@ -45,6 +45,8 @@ jobs:
 
       - name: Check
         run: ./gradlew $GRADLE_ARGS check jacocoTestReport
+             -Pnpm.publish.repository.github.authToken=batman
```
twyatt, line 48: nitpick

I do love having little "easter eggs" or fun commentary in code, but I worry that `batman` isn't immediately obvious to be a placeholder to someone who later joins the team and is diving into this code.

I'd suggest either adding a fun comment or updating the token to be clearer of intent.

```yml
      - name: Check
        # Using placeholder auth token and version for Gradle tasks that require its presence but don't use it.
        run: ./gradlew $GRADLE_ARGS check jacocoTestReport
             -Pnpm.publish.repository.github.authToken=batman
```

or...

```yml
      - name: Check
        # "Come on, Robin, to the Batcave. We haven't one moment to lose!"
        # Using placeholder auth token and version for Gradle tasks that require its presence but don't use it.
        run: ./gradlew $GRADLE_ARGS check jacocoTestReport
             -Pnpm.publish.repository.github.authToken=api-test-auth-token-placeholder
```

koap/apiTests/decoder.test.ts
```diff
@@ -0,0 +1,105 @@
...
+    expect(decodedMessage.token).toEqual(new kotlin.Long(0xCAFE))
+    expect(decodedMessage.options.modCount).toEqual(1)
+    expect(decodedMessage.payload).toEqual(new Uint8Array([0x48, 0x65, 0x6c, 0x6c, 0x6f, 0x20, 0x55, 0x44, 0x50, 0x21]))
+});
+
+test('decodeWithUdpHeader :: UDP payload :: returns header correctly', () => {
+
+    const encodedMessage = new Uint8Array([0x42, 0x01, 0xFE, 0xED, 0xCA, 0xFE, 0xB5, 0x2F, 0x74, 0x65, 0x73, 0x74, 0xFF, 0x48, 0x65, 0x6C, 0x6C, 0x6F, 0x20, 0x55, 0x44, 0x50, 0x21])
+
+    const mType = koap.Message.Udp.Type.Acknowledgement
```
twyatt, line 36: This looks like Hungarian notation, which is an [abandoned style for Android code](https://jakewharton.com/just-say-no-to-hungarian-notation/); so could be slightly confusing for long time Android devs.

```suggestion
    const messageType = koap.Message.Udp.Type.Acknowledgement
```

or

```suggestion
    const type = koap.Message.Udp.Type.Acknowledgement
```

koap/apiTests/decoder.test.ts
```diff
@@ -0,0 +1,105 @@
...
+    const method = koap.Message.Code.Method.POST
+    const token = new kotlin.Long(0xCAFE)
+    const id = 0xFEED
+    const header = new koap.Header.Udp(
+        12,
+        1,
```
twyatt, line 42: nitpick: Would be easier to read if we knew what these variables are for. They just look like [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)) as-is.
Could either add comments, or extract them as local variables.

koap/build.gradle.kts
```diff
@@ -53,3 +53,43 @@ npmPublishing {
...
+
+    description = "Builds the JS Koap package and installs it to the test suite for use"
+    group = "Verification"
+    dependsOn("assembleJsNpmPublication")
+
+    doLast {
+        val response = runCommand("cd apiTests && npm install --force file://../build/publications/npm/js")
+        println("$response")
+    }
+}
```
twyatt, line 67: We should probably switch this (and the tasks below) to [`Exec` based tasks](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.Exec.html).

Then you could drop the `cd` portion of all the commands and just define the working directory.

I _think_ something like the following would work...?

```suggestion
task<Exec>("apiTestNpmInstall") {
    description = "Builds the JS Koap package and installs it to the test suite for use"
    group = "Verification"
    dependsOn("assembleJsNpmPublication")
    workingDir("apiTests")
    commandLine("npm", "install", "--force", "file://../build/publications/npm/js")
}
```

koap/apiTests/encoder.test.ts
```diff
@@ -0,0 +1,51 @@
...
+    emptyPayload
+  )
+  const encoded = koap.encode(message)
+  
+  expect(encoded).toEqual(new Int8Array([0xA1, 0x01, 0x42, 0xB4, 0x69, 0x6E, 0x66, 0x6F, 0x44, 0x62, 0x61, 0x74, 0x74]));
+});
+
+test('encodeHeader :: UDP header :: correct encoding', () => {
+
+  const mType = koap.Message.Udp.Type.Confirmable
```
twyatt, line 30: Hungarian notation should not be used here.

.github/workflows/ci.yml
davidtaylor-juul: There is an alternative. The token can be specified in the Gradle task directly inside the `repository("github") { ... }` block: `authToken = "notarealtoken"`.

This will fail if you try to publish to GitHub for real unless it is overridden from the command line but that is fine as the CI task is the only place it happens and it does exactly that.

Given how much friction there is in having to supply a dummy token (and even having to supply it in order to see the tasks in the `./gradlew tasks` list) this may be a preferred solution.

twyatt: Oh, yes, I prefer @davidtaylor-juul's solution.

twyatt:
@Phoenix7351 can you also update the PR title and description when you get a change. Thanks!

koap/build.gradle.kts
Phoenix7351: runCommand() is not being used now. Want me to get rid of it?

twyatt: > runCommand() is not being used now. Want me to get rid of it?

Yes, please. 😄

twyatt (approved):

davidtaylor-juul (approved):

Issue #41, opened 2020-12-21
Online CoAP decoder


twyatt:
Similar to cbor.me, an online encoder/decoder tool should be created. Tool should be published to GitHub Pages on commits merged to `main`.

twyatt:
Changed scope to only include decoding of messages (we don't have much need for an online tool for encoding at this time).
Closing as decoding was implemented via #42.

PR #42, opened 2021-01-07
Setup basic webapp for decoding CoAP messages

twyatt:
Partial (decoding only) implementation for #41.

GitHub Actions updates:
- Split CI and publish/snapshot actions
- Deploy online decoder on merge to `main`

https://user-images.githubusercontent.com/98017/105253420-4b055600-5b34-11eb-94e2-bd7d2b9eaa2c.mov

---

To experiment with the webapp on your local machine, checkout branch then:

```
$ ./gradlew assembleGitHubPages
$ open build/gh-pages/index.html
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,6 +39,9 @@ jobs:
       - name: Check
         run: ./gradlew $GRADLE_ARGS check jacocoTestReport
 
+      - name: Build webapp
+        run: ./gradlew $GRADLE_ARGS assembleGitHubPages
+
       - name: Codecov
         uses: codecov/codecov-action@v1
         with:
--- a/webapp/src/main/kotlin/serialization/MessageSerializer.kt
+++ b/webapp/src/main/kotlin/serialization/MessageSerializer.kt
@@ -1,0 +1,147 @@
+@file:UseSerializers(
+    TypeSerializer::class,
+    CodeSerializer::class,
+    OptionSerializer::class,
+)
+
+package com.juul.koap.serialization
+
+import com.juul.koap.Message
+import com.juul.koap.Message.Code
+import com.juul.koap.Message.Option
+import com.juul.koap.Message.Udp.Type
+import com.juul.koap.Message.Udp.Type.Acknowledgement
+import com.juul.koap.Message.Udp.Type.Confirmable
+import com.juul.koap.Message.Udp.Type.NonConfirmable
+import com.juul.koap.Message.Udp.Type.Reset
+import kotlinx.serialization.KSerializer
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.UseSerializers
+import kotlinx.serialization.descriptors.PrimitiveKind
+import kotlinx.serialization.descriptors.PrimitiveSerialDescriptor
+import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.encoding.Encoder
+
+private abstract class MessageSurrogate {
+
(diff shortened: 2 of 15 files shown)


twyatt:
Note to self: update build scripts to deploy on merges to `main`.

webapp/src/main/resources/index.html
```diff
@@ -0,0 +1,45 @@
+<!doctype html>
+<html lang="en" xmlns="http://www.w3.org/1999/html">
+    <head>
+        <meta charset="UTF-8">
```
cedrickcooke, line 4: Fun fact this is one of my biggest pet peeves (not your fault, I realize it's the spec). How the hell is a text parser supposed to find `<meta charset="UTF-8">` if it wasn't already using the correct encoding? And why doesn't HTML close all tags `/>` 😭

.github/workflows/ci.yml
```diff
@@ -1,27 +1,14 @@
-# Each virtual machine has the same hardware resources available.
-#
-# - 2-core CPU
-# - 7 GB of RAM memory
-# - 14 GB of SSD disk space
-#
-# https://help.github.com/en/actions/automating-your-workflow-with-github-actions/virtual-environments-for-github-hosted-runners
-
 name: CI
```
cedrickcooke, line 1: Nit: we could go with a more descriptive name than just CI. "Branch Check"?

cedrickcooke (approved):

webapp/src/main/resources/index.html
twyatt: Haha. The web is a crazy-wild-beast.

.github/workflows/ci.yml
twyatt: It runs on pushes (e.g. pushing a commit to `main` or even pushes to PR branches), so, technically it is checking a branch (but more the commits on that branch).

So, "branch check" is more specific, but I think CI is so engrained into me now it feels more right here?

webapp/src/main/kotlin/serialization/MessageSerializer.kt
```diff
@@ -0,0 +1,147 @@
...
+
+    override fun deserialize(decoder: Decoder): Code = TODO("Not yet implemented")
+}
+
+private object OptionSerializer : KSerializer<Option> {
+
+    override val descriptor: SerialDescriptor = PrimitiveSerialDescriptor("Option", PrimitiveKind.STRING)
+
+    override fun serialize(encoder: Encoder, value: Option) {
+        val serialized = when (value) {
```
sdonn3, line 130: seems like the when here and below could be simplified into one?

twyatt: Tried a few things but couldn't figure out how to simplify it further?

sdonn3: might just be personal preference and thus I would never enforce it on ya but I think

`when (value) {
	Option.ContentFormat.PlainText -> "Content-Format:text/plain; charset=utf-8"
	...`

would be cleaner and easier to read?

sdonn3 (approved):

webapp/src/main/kotlin/serialization/MessageSerializer.kt
twyatt: Oh, gotchya. Ya, I agree with that.

Unfortunately I snuck some other changes in to properly handle detecting encoding by also reading the `Accept` option. It muddies up the entire thing. So it has gotten slightly messy, but just being a debug tool I think that's acceptable.

**tl;dr** code has changed enough that original comment isn't really applicable anymore.

PR #44, opened 2021-02-23
Adding Message.Code.toInt() function.

Phoenix7351:
Adding functionality to get the message's code as it was sent over Coap _before_ being decoded into 'class' and 'detail.'

### DIFF
--- a/koap/src/commonMain/kotlin/Message.kt
+++ b/koap/src/commonMain/kotlin/Message.kt
@@ -338,6 +338,8 @@ sealed class Message {
         abstract val `class`: Int
         abstract val detail: Int
 
+        fun toInt(): Int = (this.`class` shl 5) or detail
+
         /** RFC 7252: 12.1.1. Method Codes */
         sealed class Method(
             override val `class`: Int,
--- a/koap/src/commonTest/kotlin/MessageTest.kt
+++ b/koap/src/commonTest/kotlin/MessageTest.kt
@@ -62,6 +62,35 @@ class MessageTest {
         assertToString(GatewayTimeout, "GatewayTimeout")
         assertToString(ProxyingNotSupported, "ProxyingNotSupported")
     }
+
+    @Test
+    fun messageCodeToInt_validCode_correctOutputInt() {
+        assertEquals(GET.toInt(), 1)
+        assertEquals(POST.toInt(), 2)
+        assertEquals(PUT.toInt(), 3)
+        assertEquals(DELETE.toInt(), 4)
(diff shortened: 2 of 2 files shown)


koap/src/commonMain/kotlin/Message.kt
```diff
@@ -338,6 +338,8 @@ sealed class Message {
         abstract val `class`: Int
         abstract val detail: Int
 
+        fun toInt(): Int = (this.`class` shl 5) or detail
```
twyatt, line 341: What is this needed for? In other words, when do we favor an `Int` over a typed constant (e.g. `GET`, `POST`, etc)?

Phoenix7351: This pertains to the conversation we were having on Thursday. As per the gateway_types.md document you shared with me, the response of a SecureCoapTask requires:

`The response code of the actual CoAP Response generated by the CoAP request.`

twyatt: This feels like a fairly specialized need, I'd rather wait to promote it to the library when it's clear that more generally needed.

I believe this is fairly specific to our internal use case, so feels like this would be more appropriate as an extension function in our internal library that needs this functionality?

Phoenix7351: "In the RFC, these Codes are defined in the form #.## so also don't quite fit the paradigm of an Int, despite the choice to pack the components in an Int"

Yeah, that's fair. I went down the route of putting it in koap, though, based on the conversation we had on slack. I'm fine with putting it in runtime-io-client, but I might not have understood the reasons we decided to put it in here in the first place.

Phoenix7351: "Do we have the need for converting Code to an Int in multiple libraries?"

Not that I'm aware of.

twyatt: > but I might not have understood the reasons we decided to put it in here in the first place.

If the integer constants are defined in the RFC then they make sense to go in KoAP. Looking over the RFC, it looks like they define `Code` as 2 sub components (`c.dd`) whereas you have `class` and `detail`, which we already represent. So the combined `Int` representation felt out of place because it doesn't explicitly appear in the RFC (other than saying the components are packed together for storage/transmission purposes).

Phoenix7351: Right, I get that.

Phoenix7351:
This logic will be moved to runtime-io-client.

PR #51, opened 2021-06-25
JavaScript IR Support

### LINKED ISSUE #50: Support JavaScript IR Compiler

KVelasco:
Resolves #50


koap/build.gradle.kts
```diff
@@ -28,6 +27,12 @@ kotlin {
             }
         }
 
+        val jsMain by getting {
+            dependencies {
+                implementation(npm("os-browserify", "0.3.0"))
+            }
+        }
+
```
KVelasco, line 35: okio multiplatform for JavaScript uses the OS module. Since we target the browser rather than node we have to provide the polyfill module

twyatt:
Looks like build is failing due to missing `browserDevelopmentWebpack` task. Perhaps when using IR we don't get that task anymore? Wonder what the equivalent task is.

KVelasco:
@twyatt you are correct https://kotlinlang.org/docs/js-project-setup.html#building-executables. But we can not use `binaries.executable()` another due to supporting both IR and Legacy

twyatt:
Hopefully they've provided a way to bundle into a JavaScript consumable package.
For example, our sample app: https://juullabs.github.io/koap/

Uses it via:
https://github.com/JuulLabs/koap/blob/60b0e42cfb1ff8dff6f5e222ffe1d290a3a775f4/webapp/src/main/resources/index.html#L7

build.gradle.kts
```diff
@@ -7,28 +7,19 @@ buildscript {
...
-    kotlin("js") version "1.4.31" apply false
-    kotlin("plugin.serialization") version "1.4.31" apply false
+    kotlin("multiplatform") version "1.5.10" apply false
+    kotlin("js") version "1.5.10" apply false
+    kotlin("plugin.serialization") version "1.5.10" apply false
     id("org.jmailen.kotlinter") version "3.2.0" apply false
     id("com.vanniktech.maven.publish") version "0.14.0" apply false
-    id("org.jetbrains.dokka") version "1.4.30"
+    id("org.jetbrains.dokka") version "1.4.32"
     id("binary-compatibility-validator") version "0.2.3"
```
twyatt, line 16: Any chance you can bump this to 0.6.0 (assuming it doesn't break anything)?

```suggestion
    id("binary-compatibility-validator") version "0.6.0"
```

Seeing a lot of changes in `koap.api` and wondering if it is just an incompatibility of `binary-compatibility-validator` w/ the newer Kotlin compiler. Or perhaps the new compiler is showing us that we had some function with too public of visibility.

KVelasco:
@twyatt solved the executable error by having `koap` on `BOTH` and `webapp` on `Legacy` I figure there is no reason we should have `webapp` on `BOTH` since it just gets deployed to Github pages, didn't not want this to block us. Seeing a lot of polyfill errors due to upgrading to webpack 5. I could downgrade `webapp` to webpack 4 and to not even tackle this issues and that would probably be the easiest solution. what are your thoughts?

build.gradle.kts
KVelasco: I noticed that as well. Let me give this a try and see what happens

twyatt:
> @twyatt solved the executable error by having `koap` on `BOTH` and `webapp` on `Legacy` I figure there is no reason we should have `webapp` on `BOTH` since it just gets deployed to Github pages, didn't not want this to block us. Seeing a lot of polyfill errors due to upgrading to webpack 5. I could downgrade `webapp` to webpack 4 and to not even tackle this issues and that would probably be the easiest solution. what are your thoughts?

Clever w/ having the sample webapp use `LEGACY`. Totally fine w/ the sample also using an older Webpack for now.

KVelasco:
@twyatt looks like using Webpack 4 for `webapp` solves our build issues. Do we still need the `apiTests` directory? Can we delete that whole folder?

twyatt:
> @twyatt looks like using Webpack 4 for `webapp` solves our build issues. Do we still need the `apiTests` directory? Can we delete that whole folder?

I don't think we need/want `apiTests` anymore since the IR compiler does a better job of exposing the public API to JavaScript consumers (or at least that's how I understand it) but I'll let @davidtaylor-juul comment as to if we should delete it at this time (when he reviews this PR).

_I'm hoping his answer is that we don't need it because it is in version control if we ever need to get it back, but he may know of a reason to keep it and include it in the test runs for PRs._ 🤷

twyatt:
@KVelasco are you able to still run the local web server and confirm that the sample webapp still runs as expected?

davidtaylor-juul:
> > @twyatt looks like using Webpack 4 for `webapp` solves our build issues. Do we still need the `apiTests` directory? Can we delete that whole folder?
> 
> I don't think we need/want `apiTests` anymore since the IR compiler does a better job of exposing the public API to JavaScript consumers (or at least that's how I understand it) but I'll let @davidtaylor-juul comment as to if we should delete it at this time (when he reviews this PR).
> 
> _I'm hoping his answer is that we don't need it because it is in version control if we ever need to get it back, but he may know of a reason to keep it and include it in the test runs for PRs._ 🤷

We can remove `apiTests`, along with the associated Gradle tasks. This was an interim measure for validating the API when using the legacy compiler and won't be needed anymore as the IR compiler will generate the typings correctly.

davidtaylor-juul:
> @twyatt you are correct https://kotlinlang.org/docs/js-project-setup.html#building-executables. But we can not use `binaries.executable()` anymore due to supporting both IR and Legacy

Ideally we would use `binaries.library()` now for all of these repos, but I'm not sure if that works when targeting "both"?

build.gradle.kts
davidtaylor-juul: Those functions are marked `internal` unless I'm misreading that output. Perhaps adding `explicitApi()` will help?

KVelasco:
> @KVelasco are you able to still run the local web server and confirm that the sample webapp still runs as expected?
> 
> You can test with the following input: `42 01 FE ED CA FE B7 65 78 61 6D 70 6C 65` (UDP should decode, and TCP won't — if it is working).
> 
> Screen Shot 2021-06-24 at 11 50 28 PM

@twyatt I did `:webapp:browserDevelopmentRun` and got this output 

Screen Shot 2021-06-25 at 5 42 47 PM

So it looks like the webapp is still working as expected

KVelasco:
> > @twyatt you are correct https://kotlinlang.org/docs/js-project-setup.html#building-executables. But we can not use `binaries.executable()` anymore due to supporting both IR and Legacy
> 
> Ideally we would use `binaries.library()` now for all of these repos, but I'm not sure if that works when targeting "both"?

`binaries.library()` only works for `IR` but not for `BOTH`. I figured it was not worth updating the `webapp` to the `IR`  compiler since it is more trouble than it is worth. Seems like we care more about the actual `koap` library to support `BOTH` which is what this PR addresses

twyatt (approved):

KVelasco:
@twyatt updated `binary-compatibility-validator` to version `0.6.0`

davidtaylor-juul (approved):
🚀

PR #79, opened 2021-12-21
Drop `LEGACY` JavaScript compiler support

twyatt:


koap/build.gradle.kts
```diff
@@ -15,7 +15,7 @@ jacoco {
 
 kotlin {
     jvm()
-    js(BOTH).browser()
+    js().browser()
```
davidtaylor-juul, line 18: Doesn't this then default to `LEGACY`? Or did that behavior change with the latest compiler?

I guess the real question is what is the default value of `kotlin.js.compiler`?

twyatt: I assumed IR was the default, but I'll set `kotlin.js.compiler` explicitly to be sure. 👍

twyatt: It appears you were right. After setting it explicitly I'm not getting the GitHub pages to build. Investigating...

twyatt: Fixed. Apparently the distribution directory changed between the compilers.

QuantumRand (approved):

PR #123, opened 2022-06-07
Add `macosArm64` target

twyatt:


koap/build.gradle.kts
```diff
@@ -17,6 +17,7 @@ kotlin {
     jvm()
     js().browser()
     macosX64()
+    macosArm64()
```
QuantumRand, line 20: Wow, hard to believe it can be this simple when there's no platform specific source sets

QuantumRand (approved):

cedrickcooke (approved):

PR #264, opened 2024-04-28
Switch to Code Climate for code coverage reporting

twyatt:


twyatt:
Looks like we're sticking with Codecov until https://github.com/paambaati/codeclimate-action/issues/744 is resolved.

twyatt:
Upstream repo that issue was in has been archived. Since it seems Code Climate has abandoned support, we'll be abandoning this effort as well.

PR #325, opened 2025-03-03
Integrate Kover for code coverage reporting

twyatt:
"Replaces" JaCoCo with [Kover]. Although [Kover] can use JaCoCo under-the-hood, it is much simpler to configure than JaCoCo, and in newer versions of Kotlin, JaCoCo will become much more difficult to setup directly (since [`java` plugin is no longer allowed](https://github.com/JuulLabs/koap/actions/runs/13467076524/job/37634944057?pr=323) to be used) — this will unblock #323.

[Kover]: https://github.com/Kotlin/kotlinx-kover

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
           java-version: '11'
       - uses: gradle/actions/setup-gradle@v4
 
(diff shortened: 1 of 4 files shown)


cedrickcooke (approved):
❤️

Issue #299, opened 2024-10-07
Can't decode Token of length 6


anuar2k:
Per CoAP spec, the TKL field is in range of 0 to 8. At the same time, the logic which parses variable-sized integers only supports inputs of 0,1,2,3,4 and 8 bytes. I believe [this is the relevant line](https://github.com/JuulLabs/koap/blob/4da666d8a7aa179419c5d903c603814688bca86f/koap-core/src/commonMain/kotlin/Decoder.kt#L259).

This makes https://juullabs.github.io/koap/ unable to decode messages like `568400004f6273657277`, which has TKL of 6. I suppose that TKLs of 5 and 7 aren't supported, too.

twyatt:
Thanks for the report!
It will probably be a long while before we can get to this though.
Pull-requests are welcome though. ❤️

PR #329, opened 2025-03-20
Support tokens of any length

zalenskivolt:
Addresses #299

### DIFF
--- a/koap-core/src/commonMain/kotlin/ByteArrayReader.kt
+++ b/koap-core/src/commonMain/kotlin/ByteArrayReader.kt
@@ -67,6 +67,16 @@ internal class ByteArrayReader(
         return ((bytes[index++].toLong() and 0xFF) shl 24) or readUInt24().toLong()
     }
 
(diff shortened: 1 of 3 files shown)


koap-core/src/commonMain/kotlin/ByteArrayReader.kt
```diff
@@ -67,6 +67,16 @@ internal class ByteArrayReader(
         return ((bytes[index++].toLong() and 0xFF) shl 24) or readUInt24().toLong()
     }
 
+    /** Reads N-bytes to acquire a 64-bit signed int. */
```
twyatt: nitpick

```suggestion
    /** Reads [n]-bytes to acquire a 64-bit signed int. */
```

twyatt (approved):
Looks great! I pushed some commits that restructure the unit tests.
Will merge after getting another approval.

davertay-j (approved):
Excellent work, thanks @zalenskivolt

PR #333, opened 2025-03-22
Add Oscore and Edhoc options

zalenskivolt:
see [IANA CoAP Option Numbers](https://www.iana.org/assignments/core-parameters/core-parameters.xhtml#option-numbers)

\# | Option | RFC
-- | -- | --
9 | OSCORE | [[RFC8613](https://www.iana.org/go/rfc8613) OSCORE (using COSE)]
21 | EDHOC | [[RFC9668](https://www.iana.org/go/rfc9668) Ephemeral Diffie-Hellman Over COSE]

### DIFF
--- a/koap-core/src/commonMain/kotlin/Message.kt
+++ b/koap-core/src/commonMain/kotlin/Message.kt
@@ -517,20 +517,20 @@ sealed class Message {
 
         /** RFC 8613 2. OSCORE */
         data class Oscore(
-            val coseObject: ByteArray,
+            val value: ByteArray,
         ) : Option() {
             init {
-                require(coseObject.size in OSCORE_LENGTH_RANGE) {
-                    "Oscore length of ${coseObject.size} is outside allowable range of $OSCORE_LENGTH_RANGE"
+                require(value.size in OSCORE_LENGTH_RANGE) {
+                    "Oscore length of ${value.size} is outside allowable range of $OSCORE_LENGTH_RANGE"
                 }
             }
 
             override fun equals(other: Any?): Boolean =
-                this === other || (other is Oscore && coseObject.contentEquals(other.coseObject))
+                this === other || (other is Oscore && value.contentEquals(other.value))
 
-            override fun hashCode(): Int = coseObject.contentHashCode()
+            override fun hashCode(): Int = value.contentHashCode()
 
-            override fun toString(): String = "Oscore(coseObject=${coseObject.toHexString()})"
+            override fun toString(): String = "Oscore(value=${value.toHexString()})"
         }
 
         /** RFC 9668 3.1. EDHOC */
--- a/koap-core/src/commonMain/kotlin/Message.kt
+++ b/koap-core/src/commonMain/kotlin/Message.kt
@@ -296,7 +296,7 @@ sealed class Message {
 
             override fun hashCode(): Int = etag.contentHashCode()
 
-            override fun toString(): String = "ETag(etag=${etag.toHexString()}"
+            override fun toString(): String = "ETag(etag=${etag.toHexString()})"
         }
 
         /** RFC 7252 5.10.7. Location-Path and Location-Query */
@@ -336,7 +336,7 @@ sealed class Message {
 
             override fun hashCode(): Int = etag.contentHashCode()
 
-            override fun toString(): String = "IfMatch(etag=${etag.toHexString()}"
+            override fun toString(): String = "IfMatch(etag=${etag.toHexString()})"
         }
 
         /** RFC 7252 5.10.8.2. If-None-Match */
--- a/koap-core/src/commonMain/kotlin/Message.kt
+++ b/koap-core/src/commonMain/kotlin/Message.kt
@@ -44,6 +44,7 @@ private val PROXY_SCHEME_LENGTH_RANGE = 1..255
 private val SIZE_RANGE = UINT_RANGE // Size1, Size2
 private val OBSERVE_RANGE = 0..16_777_215 // 3-byte unsigned int
 internal val BLOCK_NUMBER_RANGE = 0..0xF_FF_FF
+private val OSCORE_LENGTH_RANGE = 0..255
 private val HOP_LIMIT_RANGE = 1..255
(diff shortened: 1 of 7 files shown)


twyatt:
This is great!
I'll follow up on this after #329 and #332 are merged.

twyatt:
@zalenskivolt when you have a chance, can you rebase/fix conflicts? TY!

zalenski:
> when you have a chance, can you rebase/fix conflicts? TY!

@twyatt rebased & fixed

koap-core/src/commonMain/kotlin/Message.kt
```diff
@@ -384,6 +385,29 @@ sealed class Message {
             }
         }
 
+        /** RFC 8613 2. OSCORE */
+        data class Oscore(
+            val coseObject: ByteArray,
```
twyatt, line 390: Reading through the spec, I can't decide if `value` or `coseObject` is more appropriate here.

Screenshot 2025-03-25 at 12 13 50 AM

What was your thinking for settling on `coseObject` naming?

twyatt (commented):
Amazing job on this! 💯 

I've added a unit test for "Test Vector 8", it'd be great to replicate the test for the remaining test vectors. Feel free to take that on, or I'll get to it next weekend.

koap-core/src/commonMain/kotlin/Message.kt
zalenski: > What was your thinking for settling on coseObject naming?

Hm. I thought in section [2](https://www.rfc-editor.org/rfc/rfc8613.html#section-2) it stated that the option "contains a compressed COSE object", and I just wanted a descriptive name if I could find one.

But I realize now I have probably misinterpreted this. The option just "indicates that the CoAP message is an OSCORE message", and that the **message** contains a COSE object.

> [2](https://www.rfc-editor.org/rfc/rfc8613.html#section-2).  The OSCORE Option
>
>   The OSCORE option defined in this section (see Figure 3, which
   extends "Table 4: Options" of [[RFC7252](https://www.rfc-editor.org/rfc/rfc7252)]) indicates that the CoAP
   message is an OSCORE **message** and that _it_ <b>contains a compressed COSE
   object</b> (see Sections [5](https://www.rfc-editor.org/rfc/rfc8613.html#section-5) and [6](https://www.rfc-editor.org/rfc/rfc8613.html#section-6)).  The OSCORE option is critical, safe
   to forward, part of the cache key, and not repeatable.

What is actually in the option value described in section 6.1:

> [6.1](https://www.rfc-editor.org/rfc/rfc8613.html#section-6.1).  Encoding of the OSCORE Option Value
>
>   The value of the OSCORE option SHALL contain the OSCORE flag bits,
   the 'Partial IV' parameter, the 'kid context' parameter (length and
   value), and the 'kid' parameter as follows:

So I think it should be renamed, maybe just to value.

For the block options I did some handling of the uint value to expand it to the block number, more flag, and block size. I guess a semantically aware OSCORE option implementation should also extract all those flags, 'Partial IV' parameter and 'kid context' parameter things. But maybe that could be left out for now, with just an opaque value?

zalenski: I did a take on actually handling `partialIv`, `kidContext` and `kid` in
https://github.com/zalenskivolt/koap/pull/3
learning a lot about Kotlin nullables (since both `kidContext` and `kid` can have the flag set but be empty) :)

There's reserved values for partialIv length 6 and 7, and three flags reserved for future use, so there's some potential for future updates to the standard rendering the option unparseable, and such flags/values should be handled in specific ways as outlined in [8.2](https://www.rfc-editor.org/rfc/rfc8613.html#section-8.2).  Verifying the Request …

It could have a `malformed` flag in it, or maybe a nullable partialIv and use null there to indicate the option didn't parse? I guess these complex options are a bit trickier than the simple uint/string ones in basic CoAP.

zalenski: For upgradability/backwards compatibility, kotlin just says "Avoid using data classes in your API﻿"
https://kotlinlang.org/docs/api-guidelines-backward-compatibility.html#avoid-using-data-classes-in-your-api
and here's one suggestion for mitigation: No data:
https://jakewharton.com/public-api-challenges-in-kotlin/#mitigation-no-data

but I'll leave these considerations for the library maintainers 😉

twyatt:
@zalenskivolt could you rebase/resolve conflicts on this PR again, please?

koap-core/src/commonMain/kotlin/Message.kt
twyatt: https://github.com/zalenskivolt/koap/pull/3 looks great, but let's leave that for review after we get this PR in. So ya, I appreciate you keeping it as a separate PR for now. 👍

zalenski:
> Amazing job on this! 💯
> 
> I've added a unit test for "Test Vector 8", it'd be great to replicate the test for the remaining test vectors. Feel free to take that on, or I'll get to it next weekend.

Added test vectors 4-7. (test vector 1-3 is just key derivation, no OSCORE options)

twyatt (approved):
💯 🚀

zalenski:
> Added test vectors 4-7.

…and found a bug in https://github.com/zalenskivolt/koap/pull/3 because of them! :)

twyatt:
@zalenskivolt please rebase/fix conflicts, when you have a chance. TY!

zalenski:
> @zalenskivolt please rebase/fix conflicts, when you have a chance. TY!

rebased & fixed

QuantumRand (approved):
Awesome to see expanded support for more of the CoAP spec! Thanks @zalenskivolt!

koap-core/src/commonMain/kotlin/Message.kt
zalenskivolt: > [zalenskivolt#3](https://github.com/zalenskivolt/koap/pull/3) looks great, but let's leave that for review after we get this PR in. So ya, I appreciate you keeping it as a separate PR for now. 👍

-> #343

PR #335, opened 2025-03-22
Add block-wise options

zalenskivolt:
see [IANA CoAP Option Numbers](https://www.iana.org/assignments/core-parameters/core-parameters.xhtml#option-numbers)

\# | Option | RFC
-- | -- | --
19 | Q-Block1 | [[RFC9177](https://www.iana.org/go/rfc9177) Block-Wise robust]
23 | Block2 | [[RFC7959](https://www.iana.org/go/rfc7959) Block-Wise][[RFC8323](https://www.iana.org/go/rfc8323) TLS+][[RFC8613](https://www.iana.org/go/rfc8613) OSCORE]
27 | Block1 | [[RFC7959](https://www.iana.org/go/rfc7959) Block-Wise][[RFC8323](https://www.iana.org/go/rfc8323) TLS+][[RFC8613](https://www.iana.org/go/rfc8613) OSCORE]
28 | Size2 | [[RFC7959](https://www.iana.org/go/rfc7959) Block-Wise][[RFC8613](https://www.iana.org/go/rfc8613) OSCORE]
31 | Q-Block2 | [[RFC9177](https://www.iana.org/go/rfc9177)  Block-Wise robust]
(60 | Size1) | [[RFC7252](https://www.iana.org/go/rfc7252) CoAP][[RFC8613](https://www.iana.org/go/rfc8613) OSCORE]

(Size1 was already implemented, some tests added)

TLS+ = TCP, TLS, and WebSockets


koap-core/src/commonMain/kotlin/Message.kt
```diff
@@ -356,6 +382,93 @@ sealed class Message {
...
+        data class Block1(
+            val blockNumber: Long,
+            val more: Boolean,
+            val blockSize: Int,
+        ) : Option() {
+            init {
+                val numMSzx = numMSzx(blockNumber, more, blockSize)
+                require(numMSzx in BLOCK_RANGE) {
+                    "Uri-Port value of $numMSzx is outside allowable range of $BLOCK_RANGE"
+                }
```
zalenskivolt, line 419: started with data classes with a val numMSzx, but that seems error prone and hard to use.

not sure if the companion object should be where it is, or part of Message or its own class?

twyatt:
This is great!
I'll follow up on this after #329 and #332 are merged.

twyatt:
@zalenskivolt when you have a chance, can you rebase/fix conflicts? TY!

zalenski:
> when you have a chance, can you rebase/fix conflicts? TY!

@twyatt rebased & fixed

twyatt:
@zalenskivolt rebase/fix conflicts again, please.

koap-core/src/commonMain/kotlin/Message.kt
zalenski: I squashed the old version away, since the separate blockNumber, more flag, blockSize seems much more friendly

> not sure if the companion object should be where it is, or part of Message or its own class?

especially since you can only have one companion object block, I guess it should be at the start (or end)?

should the methods be called something with "block", or should it be moved to some block utility companion object somewhere else?

twyatt:
@zalenskivolt I pushed a couple commits to restructure things a little.

I changed the size to an `enum` to enforce only specific sizes, but I'm not sure if it is the best approach (the enum naming is awkward) but it does provide better compile-time safety. 🤷 

Let me know what you think.

koap-core/src/commonMain/kotlin/Message.kt
```diff
@@ -387,6 +407,112 @@ sealed class Message {
...
+
+            @Suppress("ktlint:standard:enum-entry-name-case")
+            enum class Size(internal val intValue: Int) {
+                `16`(0),
+                `32`(1),
+                `64`(2),
+                `128`(3),
+                `256`(4),
+                `512`(5),
+                `1024`(6),
```
twyatt, line 423: I was torn on the best approach here: I don't like the ticks but prefixes felt ugly as well.
Can't decide on which is the lesser of the evils. 🙈

koap-core/src/commonTest/kotlin/DecoderTest.kt
```diff
@@ -203,7 +204,7 @@ class DecoderTest {
                 D1 0E # Option Delta: 27, Option Length: 1
                 AB    # Option Value: 3<<4 | 0x8 | 3
             """,
-            expected = Block1(10, true, 128),
+            expected = Block1(Block.Size.`128`, true, 10),
```
zalenski, line 207: I feel block number, more flag, block size reads more naturally than reversing the order.
It matches all (big endian) hex output, and the descriptions in the RFC etc.

"it's the 100:th… with more… kilobyte"!

koap-core/src/commonMain/kotlin/Message.kt
zalenski: could go for `kB` for 1024, and probably cover 90+% of use cases! 😉

zalenski: I think Size should have an easy way to access the actual block size in bytes as a number.
For BERT blocks, the basic block size stays 1kB, just that you can have more in one message.

Wouldn't `szxValue` be more clear than `intValue` (at least if you look it up in the rfc)?

koap-core/src/commonMain/kotlin/Block.kt
```diff
@@ -0,0 +1,33 @@
...
+
+private fun <T : Message.Option.Block> blockOf(type: KClass<T>, value: Long): T {
+    val number = value shr 4
+    require(number in BLOCK_NUMBER_RANGE) {
+        "Block number $number (from option value of $value) is outside allowable range of $BLOCK_NUMBER_RANGE"
+    }
+    val more = (value and 0x8L) != 0L
+    val size = (value and 0b111).toInt().let {
+        Message.Option.Block.Size.entries
+            .first { entry -> entry.intValue == it }
```
zalenskivolt, line 22: could rely/require the Sizes to be defined with ordinal == szx value, and use Block.Size.entries[value and 0b111] for lookup?

koap-core/src/commonMain/kotlin/Message.kt
zalenskivolt: using ordinal to match the szx value, they would be free to hold a single public size: Int value

koap-core/src/commonTest/kotlin/DecoderTest.kt
twyatt: Good point.
Definitely agree on the `number` going first.
Wondering though if it'd make sense to have `more` as the last parameter with a default of `false`?

koap-core/src/commonMain/kotlin/Message.kt
twyatt: > Wouldn't `szxValue` be more clear than `intValue` (at least if you look it up in the rfc)?

I agree that w.r.t. the RFC it is much clearer, but I would like KoAP to be a little more intuitive/approachable (whereas it doesn't require users to look at the RFCs). At least, that is the hope.

twyatt: > I think Size should have an easy way to access the actual block size in bytes as a number.

I've exposed the value as `size` property on `Size`. A bit redundant; assuming block sizes are always in bytes, perhaps it would make sense to name the property `bytes`?

twyatt: > I was torn on the best approach here: I don't like the ticks but prefixes felt ugly as well. Can't decide on which is the lesser of the evils. 🙈

I ended up adding a `blockSizeOf(int, ..)` function that can be optionally used and makes for nicer syntax. Not sure if I'm adding unnecessary complexity with it though. 🤷

koap-core/src/commonTest/kotlin/DecoderTest.kt
zalenski: I would lean toward rfc order and no default. Anyone using these options would need to know the details anyway, and probably set the more flags explicitly if block-wise is used. Named arguments can always be used, in any order if wanted.

koap-core/src/commonMain/kotlin/Message.kt
zalenski: > > I think Size should have an easy way to access the actual block size in bytes as a number.
> 
> I've exposed the value as `size` property on `Size`. A bit redundant; assuming block sizes are always in bytes, perhaps it would make sense to name the property `bytes`?

hm. maybe the field and type should be called `blockSize`/`BlockSize` rather than just `size`/`Size`? then I think `bytes` makes sense, and is the same as in `Size1` and `Size2`.

twyatt (approved):

sdonn3 (approved):
Very clean, tested and well documented, I love it!

PR #340, opened 2025-03-22
Add tests for Option `toString`

zalenskivolt:


cedrickcooke (approved):

twyatt:
I'm reluctant to add tests around `toString` (despite you catching / fixing some small issues). As we don't / shouldn't provide any guarantees around the format of `toString` output — it is intended for debugging/logging.

Tests just add an unnecessary maintenance burden when making changes to the output format of `toString`. Although my statement loses a bit of weight in the wake of you finding bugs in some of the `toString` outputs. 🙈 

Definitely open to changing / fixing some of the `toString` outputs, I'm just hesitant to add units tests around it.

As for the serialization support (via kotlinx serialization), your tests are exposing a lot of gaps in that support. It looks like it could use a lot of love / fixing. I feel that effort should be broken out into a separate PR, as it may warrant a bit of iteration / discussion.

zalenski:
> I'm reluctant to add tests around `toString` (despite you catching / fixing some small issues). As we don't / shouldn't provide any guarantees around the format of `toString` output — it is intended for debugging/logging.

I've updated to only test things that are actually overridden with custom code, for a lot fewer tests.
I do think those should be tested for functionality, even if there are no guarantees around them.

> Tests just add an unnecessary maintenance burden when making changes to the output format of `toString`. Although my statement loses a bit of weight in the wake of you finding bugs in some of the `toString` outputs. 🙈

now there are a lot fewer tests, and they are also not needlessly repeated for the serialization, so I think the maintenance burden is minimal. you'd only have to update if something actually changes.

zalenski:
> Definitely open to changing / fixing some of the `toString` outputs, I'm just hesitant to add units tests around it.

those are currently also fixed in the Oscore PR, where I found them by mistake when copy-pasting one of them.

I think most code warrant testing, and also dislike when there are obstacles to testing things (for example side effects like logging, time, file&network operations, database code, etc.) how well I follow that in general is another matter.

> As for the serialization support (via kotlinx serialization), your tests are exposing a lot of gaps in that support. It looks like it could use a lot of love / fixing. I feel that effort should be broken out into a separate PR, as it may warrant a bit of iteration / discussion.

What is the wanted format for those? Should all HTTP-like options look like HTTP headers? And CoAP specific things as now, or also in a header format?

That could definitely be its own PR, with all serialization tests updated in it! 😉

koap-core/src/commonTest/kotlin/MessageTest.kt
```diff
@@ -25,11 +25,52 @@ import com.juul.koap.Message.Code.Response.ServiceUnavailable
...
+        assertToString(Echo("echo".encodeToByteArray()), "Echo(value=65 63 68 6F)")
+        assertToString(RequestTag("tag".encodeToByteArray()), "RequestTag(tag=74 61 67)")
+
+        // plain name
+        assertToString(IfNoneMatch, "IfNoneMatch")
+
+        // observe
+        assertToString(Observe(Register), "Observe(value=0/Register)")
+        assertToString(Observe(Deregister), "Observe(value=1/Deregister)")
+        assertToString(Observe(1234567), "Observe(value=1234567)")
```
zalenski, line 67: I think 0 and 1 could be sent in a notification response as a serial number, so they are not necessarily Register/Deregister requests. But maybe nice to see registration requests written out?

zalenski: or an update for another PR?

koap-core/src/commonTest/kotlin/MessageTest.kt
```diff
@@ -25,11 +25,52 @@ import com.juul.koap.Message.Code.Response.ServiceUnavailable
...
+    fun optionToStringOverrides() {
+        // content format
+        assertToString(ContentFormat.JSON, "ContentFormat(JSON)")
+        assertToString(ContentFormat.CBOR, "ContentFormat(CBOR)")
+        assertToString(ContentFormat.PlainText, "ContentFormat(PlainText)")
+        assertToString(ContentFormat(20), "ContentFormat(20)")
+        assertToString(Accept(ContentFormat.JSON), "Accept(JSON)")
+
+        // hex
+        assertToString(ETag("123".encodeToByteArray()), "ETag(etag=31 32 33)")
```
twyatt, line 55: For objects that only have a single parameter, I think it makes sense to omit it from the `toString`.

```suggestion
        assertToString(ETag("123".encodeToByteArray()), "ETag(31 32 33)")
```

koap-core/src/commonTest/kotlin/MessageTest.kt
```diff
@@ -25,11 +25,52 @@ import com.juul.koap.Message.Code.Response.ServiceUnavailable
...
+        assertToString(IfMatch("123".encodeToByteArray()), "IfMatch(etag=31 32 33)")
+        assertToString(IfMatch("".encodeToByteArray()), "IfMatch(etag=)")
+        assertToString(Echo("echo".encodeToByteArray()), "Echo(value=65 63 68 6F)")
```
twyatt, line 58: It may be easier to copy/paste if the hex format was w/o spaces.
The default `HexFormat` is lowercase as well, I think that is sensible and we should probably update to align with that.
Perhaps an `0x` prefix would be helpful to identify it as hex while still making copy/paste easy.

```suggestion
        assertToString(Echo("echo".encodeToByteArray()), "Echo(0x6563686f)")
```

koap-core/src/commonTest/kotlin/MessageTest.kt
```diff
@@ -25,11 +25,52 @@ import com.juul.koap.Message.Code.Response.ServiceUnavailable
...
+import com.juul.koap.Message.Option.UriHost
+import com.juul.koap.Message.Option.UriPort
 import kotlin.test.Test
 import kotlin.test.assertEquals
 
 class MessageTest {
 
+    @Test
+    fun optionToStringOverrides() {
+        // content format
```
twyatt, line 47: IMHO, these comments don't add any value.
If for organization, I think separate test functions would be a better alternative (e.g. `fun contentFormatOptionsToString() { .. }`).

```suggestion
```

koap-core/src/commonTest/kotlin/MessageTest.kt
twyatt: Ya, I think it warrants being pulled out of this PR, only because I'm pretty torn on the best approach, so seems like we could discuss further rather than holding this PR up.

koap-core/src/commonTest/kotlin/MessageTest.kt
zalenski: this just uses `ByteArray.toHexString()` from `Debug.kt`.
in #341, it actually uses `@ExperimentalStdlibApi`, which also implements `toHexString()` but without spaces, which I also think might be better here.

> default HexFormat is lowercase as well

I've come to like uppercase hex more and more for readability. from old c there's no real default, since you choose `%x` or `%X` in printf:s yourself… 😄

koap-core/src/commonTest/kotlin/MessageTest.kt
zalenski: removed

zalenski:
> Left some comments re: changing the `toString` implementations. If preferred, I'm fine w/ deferring those changes if we just want to get these tests in and then make changes in other PRs.

I think I'd prefer that! 😄

webapp/src/jsTest/kotlin/serialization/SerializerTest.kt
```diff
@@ -1,38 +0,0 @@
...
-import com.juul.koap.Message.Option.UriPort
-import kotlinx.serialization.UseSerializers
-import kotlinx.serialization.json.Json
-import kotlin.test.assertEquals
-import kotlin.test.Test
-
-class SerializerTest {
-    @Test
-    fun optionSerializesTo() {
-        assertOptionSerializesTo(ContentFormat.PlainText, "Content-Format: text/plain; charset=utf-8")
```
zalenski: moving serializer test to #341 only, where plenty of more special serializations are added

zalenski:
> > Left some comments re: changing the `toString` implementations. If preferred, I'm fine w/ deferring those changes if we just want to get these tests in and then make changes in other PRs.
> 
> I think I'd prefer that! 😄

Updated this PR to only test Option `toString` overrides, just like methods and responses are tested. Moved all serialization tests to #341 where a lot more of those are added. No extra comments, no changes to implementations in this PR…

So please consider if you're ok with testing the code 😄

twyatt (approved):
I'd like to tweak the output of these `toString` implementations, but as discussed, that can be saved for a later PR. For now, this looks great!

PR #342, opened 2025-03-31
Output `Observe(value=0|Register)` and `Observe(value=1|Deregister)` in `toString`

zalenskivolt:


koap-core/src/commonTest/kotlin/MessageTest.kt
```diff
@@ -25,11 +25,21 @@ import com.juul.koap.Message.Code.Response.ServiceUnavailable
...
+import com.juul.koap.Message.Option.Observe.Registration.Register
 import kotlin.test.Test
 import kotlin.test.assertEquals
 
 class MessageTest {
 
+    @Test
+    fun optionToString() {
+        assertToString(Observe(Register), "Observe(value=0/Register)")
+        assertToString(Observe(Deregister), "Observe(value=1/Deregister)")
```
twyatt, line 39: I _think_ I prefer pipe `|` instead of the slash, in part because the slash could possibly be misread/confused for division (with it sitting right next to a number). 🤷 

```suggestion
        assertToString(Observe(Register), "Observe(value=0|Register)")
        assertToString(Observe(Deregister), "Observe(value=1|Deregister)")
```

koap-core/src/commonMain/kotlin/Message.kt
```diff
@@ -406,6 +406,15 @@ sealed class Message {
...
             }
+
+            override fun toString(): String {
+                val reg = when (value) {
+                    0L -> "/Register"
+                    1L -> "/Deregister"
+                    else -> ""
+                }
+                return "Observe(value=$value$reg)"
+            }
```
twyatt, line 417:
```suggestion
            override fun toString(): String = buildString {
                append("Observe(value=")
                append(value)
                when (value) {
                    0L -> append("|Register")
                    1L -> append("|Deregister")
                }
                append(')')
            }
```

twyatt (approved):

PR #343, opened 2025-03-31
Handle Oscore `partialIv`, `kidContext`, and `kid` explicitly

zalenskivolt:


twyatt:
@zalenskivolt apologies; I've been busy on other work projects, but I plan to review this soon.

koap-core/src/commonMain/kotlin/Message.kt
```diff
@@ -580,20 +580,86 @@ sealed class Message {
...
+                    (
+                        other is Oscore &&
+                            partialIv.contentEquals(other.partialIv) &&
+                            kidContext.contentEquals(other.kidContext) &&
+                            kid.contentEquals(other.kid)
+                    )
 
-            override fun hashCode(): Int = value.contentHashCode()
+            override fun hashCode(): Int {
+                var result = partialIv.hashCode()
```
cedrickcooke, line 604: Should be `contentHashCode()`

cedrickcooke (commented):
Sorry for the delays, overall great work again. One small change and then this turns into an approve

koap-core/src/commonMain/kotlin/Message.kt
zalenskivolt, line 604:
```suggestion
                var result = partialIv.contentHashCode()
```

koap-core/src/commonMain/kotlin/Message.kt
```diff
@@ -580,20 +580,86 @@ sealed class Message {
...
+                    return byteArrayOf()
+                }
+                var value = byteArrayOf(flagBits.toByte()) + partialIv
+                if (kidContext != null) {
+                    value += byteArrayOf(kidContextLength.toByte()) + kidContext
+                }
+                if (kid != null) {
+                    value += kid
+                }
+                return value
```
zalenskivolt, line 639: these are some 20+17 lines of fairly dense/complicated code, with magic flag constants, etc.
maybe they should go in their own `.kt` file to isolate the complexity, and using constants for flags and masks?
not too happy that `kidFlag` and `kidContextFlag` is a byte/int value in one place and boolean in the other, maybe they should at least be named differently… (`kidMask`/`kidContextMask`?)

koap-core/src/commonMain/kotlin/Message.kt
```diff
@@ -580,20 +580,86 @@ sealed class Message {
...
+                    val partialIv = value.sliceArray(1 until 1 + partialIvLength)
+                    val kidContextLength = if (kidContextFlag) value[1 + partialIvLength].toInt() else 0
+                    val kidContext = if (kidContextFlag) {
+                        value.sliceArray(2 + partialIvLength until 2 + partialIvLength + kidContextLength)
+                    } else {
+                        null
+                    }
+                    val kidOffset = (if (kidContextFlag) 2 else 1) + partialIvLength + kidContextLength
+                    val kid = if (kidFlag) value.sliceArray(kidOffset until value.size) else null
+                    return Oscore(partialIv, kidContext, kid)
```
zalenskivolt, line 660: these are some 20+17 lines of fairly dense/complicated code, with magic flag constants, etc.
maybe they should go in their own `.kt` file to isolate the complexity, and using constants for flags and masks?

koap-core/src/commonMain/kotlin/Message.kt
zalenskivolt: also, as mentioned in #333 

> There's reserved values for partialIv length 6 and 7, and three flags reserved for future use, so there's some potential for future updates to the standard rendering the option unparseable, and such flags/values should be handled in specific ways as outlined in [8.2](https://www.rfc-editor.org/rfc/rfc8613.html#section-8.2). Verifying the Request …
>
> It could have a malformed flag in it, or maybe a nullable partialIv and use null there to indicate the option didn't parse? I guess these complex options are a bit trickier than the simple uint/string ones in basic CoAP.

not sure how likely the option is to change, and it could just be handled by updating the code. but som drawback compared to just a binary blob is that it might start failing if updated.

zalenskivolt: > these are some 20+17 lines of fairly dense/complicated code, with magic flag constants, etc. maybe they should go in their own `.kt` file to isolate the complexity, and using constants for flags and masks? not too happy that `kidFlag` and `kidContextFlag` is a byte/int value in one place and boolean in the other, maybe they should at least be named differently… (`kidMask`/`kidContextMask`?)

added `Oscore.kt` with constants for flags, masks, and ranges, and updated to `kid[Context]FlagBit`

zalenskivolt:
Another take: instead of breaking up the option value into parts, just do that in the toString method,
and optionally construct an Oscore option from the different parts, using `Oscore.fromParts(partialIv, kidContext, kid)`

koap-core/src/commonMain/kotlin/Message.kt
```diff
@@ -603,7 +603,17 @@ sealed class Message {
 
             override fun hashCode(): Int = value.contentHashCode()
 
-            override fun toString(): String = "Oscore(value=${value.toHexString()})"
+            override fun toString(): String {
+                val parts = oscorePartsFromOptionValue(value)
```
davertay-j, line 607: `oscorePartsFromOptionValue` can throw on malformed input data. It may be surprising to a system that catches a parsing exception and then uses this `toString` method to log the failure - they won't be expecting the log invoke to `throw` which may cause a crash. I'd recommend ensuring that `toString` cannot throw as it is typically considered a safe operation.

koap-core/src/commonMain/kotlin/Oscore.kt
```diff
@@ -0,0 +1,92 @@
...
+    val kidFlagBit = if (oscore.kid != null) KID_FLAG else 0
+    val kidContextFlagBit = if (oscore.kidContext != null) KID_CONTEXT_FLAG else 0
+    val flagBits = partialIvLength or kidFlagBit or kidContextFlagBit
+    if (flagBits == 0) {
+        // If the OSCORE flag bits are all zero (0x00), the option value SHALL be empty
+        return byteArrayOf()
+    }
+    var value = byteArrayOf(flagBits.toByte()) + oscore.partialIv
+    if (oscore.kidContext != null) {
+        value += byteArrayOf(kidContextLength.toByte()) + oscore.kidContext
```
davertay-j, line 62: Might be worth a `require` length guard on `kidContextLength` here (similar to above) to ensure that it does not exceed the implicit 255 byte truncation and avoid returning a potentially corrupted result. Would be worth adding a test case for that.

davertay-j (commented):
LGTM, some minor quibbles only.

